PR #4, opened 2015-04-17
New draft API

tschuy:
This changes the API significantly.

This puts the ADD, UPDATE request bodies into the body of the request as JSON.

It also removes the ext field (why is this a thing), and makes errors only show upon error. (This is how most APIs I'm familiar with implement errors.)

It also specifies the response bodies of API requests more fully.

### DIFF
--- a/docs/source/design/draft_api.rst
+++ b/docs/source/design/draft_api.rst
@@ -25,7 +25,7 @@ Errors will be returned only when an error has occurred. They will consist
 of an errno, an error category, and an error text. The existence of the 'error'
 key indicates an error.
 
-The following error codes are suggested:
+The following error codes will exist:
 
 1. Object not found
 
@@ -88,10 +88,6 @@ a string passed to `duration`, or an invalid datetime passed to `date_worked`.
         'errno': 5,
         'text': error
     }
-
-Possible errors can include objects not found, invalid foreign keys, and
-invalid names, etc.
-
 
 
 GET Endpoints
--- a/docs/source/design/draft_api.rst
+++ b/docs/source/design/draft_api.rst
@@ -25,7 +25,22 @@ Errors will be returned only when an error has occurred. They will consist
 of an errno, an error category, and an error text. The existence of the 'error'
 key indicates an error.
 
-ex:
+The following error codes are suggested:
+
+1. Object not found
+
+.. code:: json
+    {
+        'error': "Object not found",
+        'errno': 1,
+        'text': "Invalid " + object
+    }
+
+2. Database save failure
+
+A generic catch-all for when the database fails to load. The exact error text
+will depend on exactly what framework is used, and what error it receieves when
+attempting to save.
 
 .. code:: json
 
@@ -35,8 +50,48 @@ ex:
         'text': sql_error
     }
 
+3. Invalid foreign key
+
+This error would be returned whenever a request attempts to refer to an object
+by a foreign key that does not exist.
+
+.. code:: json
+
+    {
+        'error': "Invalid foreign key",
+        'errno': 3,
+        'text': "Invalid project"
+    }
+
+4. No Name provided
+
+This error would be returned when a Name is not passed in to an /add endpoint
+that requires a name.
+
+.. code:: json
+
+    {
+        'error': "No Name provided",
+        'errno': 4,
+        'text': error
+    }
+
+5. Invalid value
+
+This error would be returned when a given value wasn't valid -- for instance,
+a string passed to `duration`, or an invalid datetime passed to `date_worked`.
+
+.. code:: json
+
+    {
+        'error': "The provided value wasn't valid",
+        'errno': 5,
+        'text': error
+    }
+
 Possible errors can include objects not found, invalid foreign keys, and
 invalid names, etc.
+
 
 
 GET Endpoints
@@ -103,8 +158,8 @@ GET Endpoints
         "activity": 2,
         "notes":"",
         "issue_uri":"https://github.com/osu-cass/whats-fresh-api/issues/56",
-        "date_worked":null,
-        "created_at":null,
+        "date_worked": 2014-04-17,
+        "created_at": 2014-04-17,
         "updated_at":null,
         "id": 1
       },
@@ -123,8 +178,8 @@ GET Endpoints
       "notes":"",
       "issue_uri":"https://github.com/osu-cass/whats-fresh-api/issues/56",
       "date_worked":null,
-      "created_at":null,
-      "updated_at":null,
+      "created_at":2014-04-17,
+      "updated_at":2014-04-17,
       "id": 1
     }
 
@@ -173,7 +228,7 @@ To add a new object, POST to */<object name>/add* with a JSON body.
       "id": 1
     }
 
-To update an existing object, POST to */<object name>/update* with a JSON body.
+To update an existing object, POST to */<object name>/<id>* with a JSON body.
 The body only needs to contain the part that is being updated.
 
 
@@ -217,7 +272,7 @@ object.
 
 The response body upon success will be empty.
 
-Upon an error, it will return a 404, with an Object Not Found error.
+Upon an error, it will return an Object Not Found error with a 404 status code.
 
 .. code:: json
 
--- a/docs/source/design/draft_api.rst
+++ b/docs/source/design/draft_api.rst
@@ -14,9 +14,10 @@ will be sent with the JSON null value.
 
 Versions
 --------
-The API will be versioned with simple integers, 1, 2, 3, ...
+The API will be versioned with the letter 'v' followed by monotonically
+increasing integers ('v1', 'v2', 'v3', etc.)
 
-ex: https://timesync.osuosl.org/1/projects
+For example: https://timesync.osuosl.org/v1/projects
 
 
 Errors
--- a/docs/source/design/draft_api.rst
+++ b/docs/source/design/draft_api.rst
@@ -8,7 +8,7 @@ Below is the api specs for the Time Sync project.
 Format
 ------
 Responses will be returned in standard JSON format. An attempt will be made to
-keep the structure simple. Https will be used for all endpoints. Null values
+keep the structure simple. HTTPS will be used for all endpoints. Null values
 will be sent with the JSON null value.
 
 
@@ -39,7 +39,7 @@ The following error codes will exist:
 2. Database save failure
 
 A generic catch-all for when the database fails to load. The exact error text
-will depend on exactly what framework is used, and what error it receieves when
+will depend on exactly what framework is used, and what error it receives when
 attempting to save.
 
 .. code:: json
--- a/docs/source/design/draft_api.rst
+++ b/docs/source/design/draft_api.rst
@@ -81,12 +81,26 @@ that requires a name.
 
 This error would be returned when a given value wasn't valid -- for instance,
 a string passed to `duration`, or an invalid datetime passed to `date_worked`.
+This field would allow the backend to inform the user what they did wrong.
 
 .. code:: json
 
     {
         'error': "The provided value wasn't valid",
         'errno': 5,
+        'text': error
+    }
+
+6. Invalid slug
+
+This error would be returned when a slug field is invalid. It would include
+which slug was used in the return text.
+
+.. code:: json
+
+    {
+        'error': "The provided slug wasn't valid",
+        'errno': 6,
         'text': error
     }
 
@@ -101,23 +115,23 @@ GET Endpoints
       {
          "uri":"https://code.osuosl.org/projects/ganeti-webmgr",
          "name":"Ganeti Web Manager",
-         "slugs":["gwm", "ganeti-webmgr"],
-         "owner": 2,
+         "slugs":[<slugs>],
+         "owner": <username>,
          "id": 1
       },
       {...},
       ...
     ]
 
-*/projects/<project_id>*
+*/projects/<slug>*
 
 .. code:: json
 
     {
        "uri":"https://code.osuosl.org/projects/ganeti-webmgr",
        "name":"Ganeti Web Manager",
-       "slugs":["gwm", "ganeti-webmgr"],
-       "owner": 2,
+       "slugs":[<slugs>],
+       "owner": <username>,
        "id": 1
     }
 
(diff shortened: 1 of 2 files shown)


tschuy:
I drafted this API after discussion with @iankronquist and @dean . Putting the request as JSON in the request body means that JSON is common across both requests and responses, which is nice.

I would appreciate input on why we have the ext parameter in the other draft. To me, if we add something to ext, we might as well either add it to the API directly or re-version. Adding a new key to an object does not effect parsing of the other keys in the object, after all.

Regarding the error block, I'm not aware of any other API that has an error block other than of course whats-fresh-api and working-waterfronts-api. I'm not opposed to including it more than confused as to why we include it everywhere. In every language I know it's trivial to see if a key exists in a parsed JSON object -- 

```
if (object.error !== undefined ) { // do stuff; }

if request.get(error, None) != None:
     # do stuff
```

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,228 @@
...
+
+.. code:: json
+
+    {
+        'error': "Database save failed",
+        'errno': 2,
+        'text': sql_error
+    }
+
+Possible errors can include objects not found, invalid foreign keys, and
```
iankronquist: I would like to explicitly enumerate the errors.

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,228 @@
...
+-------------
+*/projects*
+
+.. code:: json
+
+    [
+      {
+         "uri":"https://code.osuosl.org/projects/ganeti-webmgr",
+         "name":"Ganeti Web Manager",
+         "slug":"gwm",
```
iankronquist: If a project has many slugs, why wouldn't they show up here?
Perhaps we want:
```json
{
  "uri":"https://code.osuosl.org/projects/ganeti-webmgr",
  "name":"Ganeti Web Manager",
  "slug":["gwm", "ganeti_webmgr", "ganeti_web_manager"],
  ...
}
```

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,228 @@
...
+    ]
+
+*/projects/<project_id>*
+
+.. code:: json
+
+    {
+       "uri":"https://code.osuosl.org/projects/ganeti-webmgr",
+       "name":"Ganeti Web Manager",
+       "slug":"gwm",
```
iankronquist: Above comment applies here too.

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,228 @@
...
+    [
+      {
+        "duration":12,
+        "user": 2,
+        "project": 3,
+        "activity": 2,
+        "notes":"",
+        "issue_uri":"https://github.com/osu-cass/whats-fresh-api/issues/56",
+        "date_worked":null,
+        "created_at":null,
```
iankronquist: This field should never be null.

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,228 @@
...
+
+.. code:: json
+
+    {
+       "name":"Documentation",
+       "slug":"doc",
+       "id": 1
+    }
+
+*/time*
```
iankronquist: Why is it "activities" and "projects" but not "times"?

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,228 @@
...
+      "created_at":null,
+      "updated_at":null,
+      "id": 1
+    }
+
+To update an existing object, POST to */<object name>/update* with a JSON body.
+The body only needs to contain the part that is being updated.
+
+
+*/projects/1*
```
iankronquist: Do mean `/projects/update`?

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,228 @@
...
+DELETE Endpoints
+----------------
+
+A DELETE request sent to any object's endpoint will result in a DELETE of the
+object. For instance, DELETE-ing /activities/1 will return a 200 and delete the
+object.
+
+The response body upon success will be empty.
+
+Upon an error, it will return a 404, with an Object Not Found error.
```
iankronquist: Let's be more specific. "If the object is not found, the response will have a 404 status code and an Object Not Found Error."

This way we can return 500 if the DB isn't connected, or 400 if they put a string in the `date_worked` field.

docs/source/design/draft_models.rst
```diff
@@ -1,4 +1,48 @@
 .. _draft-models:
 
-Models
-======
+Draft Data Models
```
iankronquist: Nit: Are these really models? This looks more like a schema.

docs/source/design/draft_models.rst
```diff
@@ -1,4 +1,48 @@
...
+=================
+
+Below are the database models to be used in TimeSync's api.
+
+Create/Modify Projects:
```
iankronquist: Why is "Create/Modify" here? Why isn't it just "Projects"?

docs/source/design/draft_models.rst
```diff
@@ -1,4 +1,48 @@
...
+-----------------------
+
+.. code-block:: python
```
iankronquist: We should specify which fields are required. I suggest the following:

Required fields:
* Projects
  - id, primary key
  - name, key
* Activities
  - id, pk
  - name, key
* Time Entry:
  - id, pk
  - project_name
  - duration
  - user
  - date

docs/source/design/draft_models.rst
```diff
@@ -1,4 +1,48 @@
...
+Create/Modify Activity:
+-----------------------
+
+.. code-block:: python
+
+    id      (int)   (immutable)
+    name    (str)
+    slug    (str)
+
+Submit Time:
```
iankronquist: Perhaps, "Time Entry"?

docs/source/design/draft_models.rst
```diff
@@ -1,4 +1,48 @@
...
+.. code-block:: python
+
+    id              auto increment    (int)
+    project_name:   slug/fuzzy        (str)
+    duration:       minutes           (int)
+    user:           username          (str)
+    activity:       slug              (str/null)
+    notes:          notes             (str)
+    issue_uri:      uri               (str)
+    date:           iso date          (str)
```
iankronquist: Perhaps "date_submitted"?

docs/source/design/draft_models.rst
```diff
@@ -1,4 +1,48 @@
...
+
+User:
+-----
+
+.. code-block:: python
+
+    # This is currently unplanned. We will address this in a future meeting.
+    id  auto-increment  (int)
```
iankronquist: We will need a username.

docs/source/design/draft_api.rst
tschuy: Can projects have multiple slugs?

docs/source/design/draft_api.rst
tschuy: Time can be an uncountable noun: https://en.wiktionary.org/wiki/time#Noun

And because of that, I think, to me, it feels more "right" to have it singular. However, I'd love to get other people's feedback on that.

docs/source/design/draft_api.rst
tschuy: No, I propose we POST to the id endpoint. The docs above were incorrect.

docs/source/design/draft_api.rst
tschuy: I agree on the rewording, and on the 500 bit, but how would they put a `date_worked` in a DELETE request?

docs/source/design/draft_models.rst
tschuy: This part was in @ElijahCaine 's branch that I branched off of. I'll update this part, however.

docs/source/design/draft_api.rst
iankronquist: I thought that you wanted to be able to refer to a project by many names. See Eli's comment above about "pgd/protein_geometry_database/pgd_core".

docs/source/design/draft_models.rst
tschuy: It's even more schema-like now, but before it was essentially the API (listed slugs as how we were storing the user/activities/projects on time entries).

docs/source/design/draft_api.rst
iankronquist: I think in this case "time" is intended to be short for "time entries". Personally, I would put "entries" or "checkins". Time can mean many things, it could be the amount of time worked, the time at which work began, or the time at which work was finished. Time is vague.
"entries" is just as bad -- what are you entering? I would use "checkins". It seems very clear to me what properties a "checkin" would have, but not what properties a "time" would have. We also "!check in" to work on IRC already.

docs/source/design/draft_api.rst
iankronquist: I agree with this idea. In that case, fix the above docs.

tschuy:
We'll need to handle multiple slugs being stored in the database somehow -- the backend verification that we're not duplicating slugs just got three times harder.

Unless we do slug_1, slug_2, and slug_3. 

But that's bad. Gross.

docs/source/design/draft_api.rst
iankronquist: Good point.

iankronquist:
Do we need the file "docs/source/design/draft_error.rst" since we enumerate errors in the docs?

docs/source/design/draft_api.rst
tschuy: Oh, I like checkins.

tschuy:
No, I don't think so. I'll get rid of that.

iankronquist:
@tschuy It sounds like we have a one to many relationship between projects and slugs, and possible activities and slugs. The project slugs will go in a table called "project_slugs", or something. This table will have two columns, a foreign key pointing at a project and a primary key which is the slug. This way slugs cannot be accidentally duplicated.

tschuy:
ah, duh, of course. Sorry, for some reason I didn't get my head around that instantly.

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,283 @@
...
+ex: https://timesync.osuosl.org/1/projects
+
+
+Errors
+------
+Errors will be returned only when an error has occurred. They will consist
+of an errno, an error category, and an error text. The existence of the 'error'
+key indicates an error.
+
+The following error codes are suggested:
```
iankronquist: We'll have to revise this when the docs are accepted. Let's just change this now.

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,283 @@
...
+
+.. code:: json
+
+    {
+        'error': "The provided value wasn't valid",
+        'errno': 5,
+        'text': error
+    }
+
+Possible errors can include objects not found, invalid foreign keys, and
```
iankronquist: nix

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,279 @@
...
+    {
+        'error': "Object not found",
+        'errno': 1,
+        'text': "Invalid " + object
+    }
+
+2. Database save failure
+
+A generic catch-all for when the database fails to load. The exact error text
+will depend on exactly what framework is used, and what error it receieves when
```
iankronquist: *receives

iankronquist:
+1 contingent on running spellcheck.

tschuy:
(That last commit message rhymes.)

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,279 @@
...
+Responses will be returned in standard JSON format. An attempt will be made to
+keep the structure simple. HTTPS will be used for all endpoints. Null values
+will be sent with the JSON null value.
+
+
+Versions
+--------
+The API will be versioned with simple integers, 1, 2, 3, ...
+
+ex: https://timesync.osuosl.org/1/projects
```
bramwelt: It's standard to preface this with a `v`: so `v1, v2, v3`.

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,279 @@
...
+-------------
+*/projects*
+
+.. code:: json
+
+    [
+      {
+         "uri":"https://code.osuosl.org/projects/ganeti-webmgr",
+         "name":"Ganeti Web Manager",
+         "slugs":["gwm", "ganeti-webmgr"],
```
bramwelt: We're not a newspaper. We should use the term 'alias' instead of 'slug'.

iankronquist: sad_slug_d__by_lillyamongthethorns

iankronquist: @Kennric expressed the opinion that we should keep 'slug' because everyone knows what a slug is, besides being a slimy bug.

tschuy:
@Kennric

docs/source/design/draft_api.rst
dean: I'm inclined to agree with @bramwelt. My 2c

tschuy:
Personally I'm of the opinion that they should be slugs.

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,280 @@
...
+         "owner": 2,
+         "id": 1
+      },
+      {...},
+      ...
+    ]
+
+*/projects/<project_id>*
```
Kennric: What is project_id? Can it be a slug? If we use slugs to make entry via commit message or other easier, being able to talk to the api in terms of slugs could be handy.

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,280 @@
...
+    [
+        {
+           "name":"Documentation",
+           "slugs":["doc"],
+           "id": 1
+        },
+        {...}
+    ]
+
+*/activities/id*
```
Kennric: Activity id? Maybe use <activity_id> in docs to make clear

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,280 @@
...
+        "issue_uri":"https://github.com/osu-cass/whats-fresh-api/issues/56",
+        "date_worked": 2014-04-17,
+        "created_at": 2014-04-17,
+        "updated_at":null,
+        "id": 1
+      },
+      {...}
+    ]
+
+*/time/id*
```
Kennric: project id? user id? id of a time entry?

docs/source/design/draft_api.rst
```diff
@@ -1,4 +1,280 @@
...
+
+Versions
+--------
+The API will be versioned with the letter 'v' followed by monotonically
+increasing integers ('v1', 'v2', 'v3', etc.)
+
+For example: https://timesync.osuosl.org/v1/projects
+
+
+Errors
```
Kennric: If we validate fields for the add interface, it might be nice to return a missing field, or incomplete data kind of error.

Kennric:
This looks ok so far, aside from being incomplete and needing a bit of doc cleanup for clarity (see notes).

tschuy:
``16:02 < kennric> well, assuming going all-slug doesn't bit us, I think that's all reasonable``

PR #17, opened 2015-07-08
added some README docs and venv/ to .gitignore

ElijahCaine:


tschuy:
:+1:

iankronquist:
Nit: I'd be a little more specific and explicitly say "run `make html` in the docs folder" or even "run `cd docs && make html`". I'm a very lazy reader though.

+1 regardless.

PR #19, opened 2015-07-09
Set line limit to 80 characters

tschuy:


src/checkins.js
```diff
@@ -52,11 +52,13 @@ module.exports = function(app) {
...
                     for (i = 0, len = slugs.length; i < len; i++) {
-                        id_activity_map[slugs[i].activity].slugs.push(slugs[i].name);
+                        id_activity_map[slugs[i].activity].slugs.push(
+                            slugs[i].name);
                     }
 
                     for (i = 0, len = times.length; i < len; i++) {
-                        times[i].activity = id_activity_map[times[i].activity].slugs;
+                        times[i].activity = id_activity_map[times[i].activity]
+                            .slugs;
```
tschuy: This is the part I like the least, followed by the parts where we need to split strings across lines.

ElijahCaine:
The present formatting doesn't please me as much as how it used to look, but that feeling is over-ruled by my love of the 80 character line limit standard.

+1

iankronquist:
It's arbitrary and we should draw a line somewhere, so it might as well be where everyone else draws it. +1.

PR #52, opened 2015-07-20
Make tests actually check for equality

tschuy:
``expect(a === b)`` is meaningless. What we really want, in order to test that equality, is ``expect(a).to.be(b)``: https://github.com/Automattic/expect.js#api


iankronquist:
r+. Looks good, tests pass.

Kennric:
Correct. +1

PR #67, opened 2015-07-28
checkProject helper function

matthewrsj:


matthewrsj:
Three tests added for check_project helper function. Needs to be reviewed @tschuy @Kennric @ElijahCaine

tests/helpers.js
```diff
@@ -0,0 +1,20 @@
+var helper = require('../src/helpers');
+
+module.exports = function(expect) {
+    describe('check projects', function() {
+        it('should return true', function(done) {
+            expect(helper.check_project('gwm')).to.equal(true);
+            done();
+        });
+
+        it('should return false', function(done) {
```
iankronquist: Can you make these more descriptive?

iankronquist: Also, you probably shouldn't use should:
http://betterspecs.org/#should

matthewrsj: @iankronquist do you mean something like

```it('should return false when passed invalid slug', ...``` ?

File: tests/helpers.js [hunk 2]
```diff
@@ -0,0 +1,20 @@
+var helper = require('../src/helpers');
+
+module.exports = function(expect) {
+    describe('check projects', function() {
```
tschuy: this name should describe what the tests do do someone reading this on the command line. Perhaps give it the exact name of the helper being tested?

src/helpers.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+
+    check_project: function(name) {
```
tschuy: Functions in javascript should, as much as I hate to say it, camelCase.

tests/test.js
```diff
@@ -55,3 +55,15 @@ describe('Endpoints', function() {
     require('./projects')(expect, request, base_url);
 
 });
+
+describe('Helpers', function() {
+
+    before(function(done) {
+        knex.migrate.latest().then(function() {
```
tschuy: You'll need to load the fixtures in this describe as well, if you want to use the fixture's data to test the helpers.

tests/helpers.js
tschuy: ``it('returns false when passed invalid slug'``, but yes.

src/helpers.js
iankronquist: We should put this in the docs and lint for it then.
https://github.com/osuosl/docs/blob/elijah/style_guides/source/development/style_guides/javascript.rst

tschuy: it's sort of in the docs: https://github.com/airbnb/javascript/blob/0aa8d97f5ba09cab807d7b5d91a8d5b2f6aad096/README.md#22.2

tschuy: I'm making a PR with it in the jscsrc. For now, this should be camelCase and I'll fix the rest.

tschuy:
Looks good to me. Anyone else want to give input before it's implemented?

tschuy:
Realized a few hours ago that we're going to have a pretty big problem with this -- this function will need to be asynchronous. Oops.

So the tests need to be rewritten for async.

tschuy:
Closing since we need to rewrite these as async. I've opened #87

PR #69, opened 2015-07-28
camelCase allTheThings

tschuy:
Because it's in the AirBnB guide, which is our guide.


iankronquist:
Is there anything else we should be adding to jscs or jshint? Now's the time to do it.

Kennric:
This looks good as-is, if there are not further needed changes to the linter, I say +1

tschuy:
None that I know of. For now.

Issue #62, opened 2015-07-27
Add a password and activity field to the model


ElijahCaine:

MaraJade:
Add password and ACTIVE fields to the user table.

PR #72, opened 2015-07-29
Modified user test to contain password and active fields

MaraJade:
issue #62

### DIFF
--- a/tests/users.js
+++ b/tests/users.js
@@ -1,16 +1,20 @@
 module.exports = function(expect, request, baseUrl) {
     describe('GET /users', function() {
-        it('should return all users in the database', function(done) {
+        it('returns all users in the database', function(done) {
             request.get(baseUrl + 'users', function(err, res) {
                 var bodyAsString = String.fromCharCode.apply(null, res.body);
                 var expectedResults = [
                     {
                         id: 1,
-                        username: 'deanj'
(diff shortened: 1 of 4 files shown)


tests/users.js
```diff
@@ -1,16 +1,20 @@
...
+        it('returns all users in the database', function(done) {
             request.get(base_url + 'users', function(err, res) {
                 var bodyAsString = String.fromCharCode.apply(null, res.body);
                 var expected_results = [
                     {
                         id: 1,
-                        username: 'deanj'
+                        username: 'deanj',
+                        active: true,
+                        password: 'pass'
```
tschuy: These should be salted bcrypt values. I can show you how to generate a few.

tschuy:
jscs is failing. Looks like your line is too long. You can disable individual jscs rules for a block with
```
//jscs:disable maximumLineLength
var long_name = 'this is a super long var that is longer than 80 characters, which is bad but if it's a long string that's maybe 90 it's better than splitting, in my opinion';
//jscs:enable maximumLineLength

MaraJade:
Should be implemented correctly now.

tschuy:
Looks good to me.

Kennric:
+1

tschuy:
+1

PR #73, opened 2015-07-29
Minor refactor of tests

tschuy:


tests/test.js
```diff
@@ -12,7 +12,35 @@ var knex = require('knex')(knexfile.mocha);
...
+                        knex('projectslugs').del().then(function() {
+                            sqlFixtures.destroy().then(function() {
+                                done();
+                            });
+                        });
+                    });
+                });
+            });
+        });
+    });
```
iankronquist: Is there a better way to do this? This is the definition of callback hell.

iankronquist: https://github.com/petkaantonov/bluebird/blob/master/API.md#all---promise
I think this is what we want.

tschuy: I don't think so. sqlite doesn't do multiple deletes at once, and a raw knex appeared to pass, but then gave a SQLITE_CONSTRAINT: UNIQUE error.

tschuy:
https://pastebin.osuosl.org/29976/

iankronquist:
For the future, these requests need to be made serially, not in parallel because sqlite isn't a real database. The Promise.all approach does not work, and chaining 'then' calls is just as ugly as what's there right now.

ElijahCaine:
That's very weird that the tests failed. They passed last night before I pushed. I'll revert the last commit and we can move on.

iankronquist:
@ElijahCaine I think we need to specify node v0.12 in the .travis.yml.

ElijahCaine:
@iankronquist 
TravisCI
```
$ nvm install 0.10

######################################################################## 100.0%

Now using node v0.10.40

$ node --version

v0.10.40
```
I think you're right :smile:

tschuy:
I had been meaning to get around to fixing that. My bad!

iankronquist:
@ElijahCaine any reason you aren't doing this?
```javascript
// databsePromises gets iterated over by Promise,
// this avoids the classic nested callback hell.
var clearDatabasePromises = [
knex('projects').del(),
knex('activities').del(),
knex('users').del(),
knex('times').del(),
knex('activityslugs').del(),
knex('projectslugs').del(),
sqlFixtures.destroy(),
];



```

ElijahCaine:
@iankronquist Can you clarify your concerns about the changes I made? We can revert them, I want to make sure it's necessary to do so.

tests/test.js
```diff
@@ -21,13 +52,7 @@ describe('Endpoints', function() {
...
-        // Clear SQLite indexes
-        knex.raw('delete from sqlite_sequence').then(function() {
-            sqlFixtures.create(knexfile.mocha, testData).then(function() {
-                done();
-            });
-        });
+        reloadFixtures(done);
     });
 
     afterEach(function(done) {
```
tschuy: This should just be clearDatabase();

tschuy:
Looks good to me. Get another review and let's get this thing in!

iankronquist:
Looks good. Tests pass, linter, etc. Our coverage is dropping, but that's no fault of this pr.

PR #74, opened 2015-07-29
Login method

tschuy:
This adds the passport middleware, which is a great little tool for checking logins. Right now it implements a broken LocalStrategy using passport-local.

To tell LocalStrategy that the user authenticated successfully (and thus test that the success test will succeed if the user is actually logged in/LocalStrategy is implemented correctly), you can put ``done(null, {username: 'tschuy'})`` on line 6 of ``src/auth/local.js``.

I'll add a user-does-not-exist and an invalid password test to this.


package.json
```diff
@@ -31,10 +31,14 @@
   "dependencies": {
     "bcrypt": "^0.8.4",
     "body-parser": "^1.12.3",
+    "chai-passport-strategy": "^0.2.0",
+    "chai": "^3.2.0",
     "expect.js": "^0.3.1",
```
iankronquist: Do we need both chai and expect.js assertion libraries? They accomplish the same task.What is the ccost of refactoring away expect.js?

tests/login.js
```diff
@@ -0,0 +1,63 @@
...
+        it('returns user for good user/pass', function(done) {
+            chai.passport.use(localPassport)
+            .success(function(user) {
+                expect(user.username).to.equal('tschuy');
+                done();
+            })
+            .req(function(req) {
+                req.body = {};
+                req.body.username = 'tschuy';
+                req.body.password = 'password';
```
lucywyman: Is this login valid?  Does 'tschuy' have permissions on the system? Is the secure?  Could someone log in with these credentials?  I haven't used passport, so it may be secure -- but 'password' on github always sets off little alarms.

tschuy: This is a valid login, stored in the fixtures. I should add a comment to the fixtures if that's possible to indicate what the user's passwords are.

what do you mean by is it secure? Passport is a middleware, and using it we can define what happens on a bad login -- redirect to a login page, send a 401, etc.

lucywyman: I get that, but shouldn't you store the valid username and password ('tschuy', 'password') in a local configuration file so they aren't publicly accessible on github?  Or is the user tschuy a dummy user who has no permissions to do anything in timesync?

tschuy: It's a test file, they're just test fixtures. This login won't do anything on a production deploy, unless someone imported the test fixtures (which would be a really bad idea).

lucywyman: Derp, sweet, ty.

tschuy: np, it's good to always keep credential safety on the mind :smile:

tschuy:
I removed expect.js from here because (1) we essentially need to use Chai to test passport, and (2) Chai has expect-style test assertions.

Chai docs: http://chaijs.com/api/bdd/

src/users.js
```diff
@@ -8,8 +8,8 @@ module.exports = function(app) {
     });
 
     app.post(app.get('version') + '/users/add', function(req, res) {
-        knex('users').insert({username: req.body.username}).then(
-                function(users) {
+        knex('users').insert({username: req.body.username})
+        .then(function(users) {
```
iankronquist: Nit: do we indent two spaces here? That's what I've seen elsewhere.

src/users.js
```diff
@@ -8,8 +8,8 @@ module.exports = function(app) {
...
             console.log(users);
```
iankronquist: Debug info crept in.

tests/login.js
iankronquist: @tschuy FYI you can't comment in the fixtures because the fixtures are json and you can't comment in json. json--

src/users.js
tschuy: the entire ``src/users.js`` will be scrapped. It's less debug info crept in, and more I accidentally changed the indentation.

src/users.js
tschuy: We could indent it two, four, zero, de-indent to the back of the line as far as jscs is concerned. We should define this somewhere. I personally like the look of no indent:

none

We could do two:

two

Or even eight: 

eight

iankronquist: It just looked funny. I don't really care, just pick one and make jscs enforce it.

tschuy: I think their internal JS parser makes that into one line. It doesn't annoy me about the indentation at ANY level, including putting it at the start of the line.

matthewrsj: I like 0 or 8, 2 spaces is too close to the 4 space indent on the next line.

ElijahCaine: I like 0 best.

tschuy: We should wait until we move the project to 2-spaces before we decide. It will probably change what we're thinking.

tschuy:
@osuosl/devs @Kennric this should be good to review.

src/routes.js
```diff
@@ -6,4 +8,15 @@ module.exports = function(app) {
     app.post(app.get('version') + '/', function(req, res) {
         res.send('hello javascript');
     });
+
+    app.post('/login',
+      passport.authenticate('local'),
+      function(req, res) {
+          console.log(req.user);
```
iankronquist: Is this here on purpose?

tschuy: I mean, the entire src/routes.js file is temporary, like src/users.js. It's a nice function to try POSTing with your favorite web development toolkit to verify things are working, and it's an example of how to use passport middleware.

Kennric:
This is looking pretty good, so far as I uderstand passport, pleae get another dev to review as well. +1

Kennric:
I don't see any major issues, but passport is a bit new to me. I'd like to see some documentation about how we are using it, with a pointer back to some useful passport docs.

ElijahCaine:
I looked through all of the code and ran the tests. The tests pass and everything seems to make sense / solves the problem it's supposed to.

I agree with @Kennric's comment about adding docs about Passport. Including somewhere in the docs a short section about Passport, that we use it, and a link to Passport's docs would be very useful.

Kennric:
Docs are a big improvement. Docs build, links all work, code still looks good. +1

ElijahCaine:
The docs are very good. I had a few comments about slight improvements; they are very good, just a few tweaks that will make them even better. Once those are added I think I can give a +1.

ElijahCaine:
Looks good to me. +1

Kennric:
+1

PR #75, opened 2015-07-30
Thai/account creation

athai:


tschuy:
Can you add documentation on how to use this?

tschuy:
I don't particularly want my password to be printed to the screen when I type it in. does prompt.js have a facility to hide this?

It should handle duplicate users cleanly instead of printing an Unhandled rejection Error.

athai:
Yeah, the password can be hidden during the prompts.  Do you not want the password to print to the screen after creating the account either?

Also, what would be added to the documentation other than:
run "npm run create-account" to start the script

tschuy:
Passwords should never be displayed, ever.

And mostly that, maybe with a nice sentence saying "there's a script included to create users".

tschuy:
Can you rebase or merge develop? And the script should probably go in ``scripts/``, not ``src/``.

scripts/create-account.js
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env node
+
+//To start the script, run 'npm run create-account'
+
+//Library requirements
```
tschuy: One space after //s before writing comments.

src/create-account.js
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env node
```
tschuy: Remove this file

scripts/create-account.js
```diff
@@ -0,0 +1,70 @@
...
+                })
+            //Exits the process after storing user info in the db
+            .then(
+                function() {
+                    process.exit(0);
+                });
+        });
+    });
+
+    console.log('\nCommand-line input received:\n');
```
tschuy: Can you just get rid of this section entirely? I feel like printing the username isn't particularly useful, especially because it's shown five lines above where i typed it.

scripts/create-account.js
```diff
@@ -0,0 +1,70 @@
...
+            .then(
+                function() {
+                    console.log('\nUser successfully created\n');
+                })
+            /*If the user enters a duplicate name, information will not be
+              added to the database. An error message will print to the screen
+              and the process will exit. */
+            .catch(
+                function() {
+                    console.log('\nINVALID ENTRY: That username is already' +
```
tschuy: We'll want to make sure to separate "I don't know what went wrong" from "the username is already taken".

tschuy: Knex passes an error variable to the catch function. If you add it as a paramter to your function, you can check the result code of the error (err.errno) against SQLite's Result Codes list:

https://www.sqlite.org/c3ref/c_abort.html

tschuy: Though I'd also add a note saying that this needs to be fixed to work on databases other than sqlite.

scripts/create-account.js
```diff
@@ -0,0 +1,84 @@
...
+                    // to work on databases other than sqlite
+
+                    /* If the error that occurs is not a constraint violation,
+                       an error message will print to the screen with a link to
+                       a list of result codes. The errno is also printed to the
+                       screen. */
+                    else {
+                        console.log('Something went wrong! Check out ' +
+                                    'https://www.sqlite.org/c3ref/' +
+                                    'c_abort.html to figure out what' +
```
ElijahCaine: Add a space after ``what``.
```
$ npm run create-account

> timesync@0.0.0 create-account /path/to/timesync-node
> node ./scripts/create-account.js

prompt: name:  user
prompt: password:
Something went wrong! Check out https://www.sqlite.org/c3ref/c_abort.html to figure out whathappened.
  Your error number is:
1
```

src/create-account.js
iankronquist: In the future you should use `git mv src dest`. It moves and adds the file.

docs/source/dev/usage.rst
```diff
@@ -0,0 +1,16 @@
...
+Usage Guide
+===========
+
+Creating a TimeSync Account
+---------------------------
+
+The TimeSync API includes a handy little script to create users. To start it
+up, run::
+
+    npm run create-account
```
ElijahCaine: We should add a little section about if you get this error:
```
$ npm run create-account
                                                                                                          
> timesync@0.0.0 create-account /path/to/timesync-node
> node ./scripts/create-account.js

prompt: name:  username
prompt: password:
Something went wrong! Check out https://www.sqlite.org/c3ref/c_abort.html to figure out what happened.

  Your error number is:
1
```
You probably need to run
```
$ npm run migrations
```
that solved the problem for me anyway.

Really some docs on expected output and how to fix basic problems would be nice.

docs/source/dev/usage.rst
```diff
@@ -0,0 +1,83 @@
...
+up, run::
+
+    $ npm run create-account
+
+This will prompt for a username and a password and store the information into
+the database.
+
+
+Expected Outputs
+----------------
```
ElijahCaine: This should be a subsection of the ``Creating a TimeSync Account``. This is done by replacing the ``---`` underlining with ``~~~``s.

docs/source/dev/usage.rst
```diff
@@ -0,0 +1,83 @@
...
+
+    > timesync@0.0.0 create-account /home/thai/projects/timesync-node
+    > node ./scripts/create-account.js
+
+    prompt: name:  username
+    prompt: password:  
+
+    User successfully created
+
+However if a username is unavailable, you'll be met with::
```
ElijahCaine: I would separate the errors into their own subsection. This makes searching and linking easier.

docs/source/dev/usage.rst
```diff
@@ -0,0 +1,93 @@
...
+    prompt: password:  
+
+    Something went wrong! Check out https://www.sqlite.org/c3ref/c_abort.html to figure out what happened.
+
+    Your error number is: 1
+
+    Exiting...
+
+According to SQlite's `result code definitions`_, an errno of 1 is indicative
+of either a SQL error or missing database. Odds are (with this script), it's
```
tschuy: "odds are, if you're running this in a development environment, it's the latter."

scripts/create-account.js
```diff
@@ -0,0 +1,84 @@
...
+var db = process.env.DATABASE || 'development';
+
+// Load the database (default = development)
+var knex = require('knex')(knexfile[db]);
+
+// Prompt user for their information
+var info = {
+    properties: {
+        name: {
+            // Note to self: '\w' matches all alphanumeric char and '_'
```
tschuy: Note, not note to self.

tschuy:
Other than that, it looks good for a development account creation tool. We'll definitely want to generalize it in the future, though.

athai:
@tschuy

iankronquist:
You need to add `#!/usr/bin/env node` if you're going to make the script executable.

```
(timesync)iankronquist@puppettop:(timesync)(thai/account-creation) → ./scripts/create-account.js 
./scripts/create-account.js: line 1: //: is a directory
./scripts/create-account.js: line 3: //: is a directory
./scripts/create-account.js: line 4: syntax error near unexpected token `('
./scripts/create-account.js: line 4: `var prompt = require('prompt');'
```
I put `test` for both the username and password, and this is what I got. I forgot to run migrations. Maybe you should remind me to delete the database and run migrations.
```
node ./scripts/create-account.js
prompt: name:  test
prompt: password:  

Something went wrong! Check out https://www.sqlite.org/c3ref/c_abort.html to figure out what happened.

  Your error number is: 1

  Exiting...


```
I think you should print usage information if someone runs `node ./scripts/create-account.js --help` or `node ./scripts/create-account.js -h`.

Good work!

tschuy:
@iankronquist read the usage docs :)

scripts/create-account.js
```diff
@@ -0,0 +1,84 @@
...
+                                    'https://www.sqlite.org/c3ref/' +
+                                    'c_abort.html to figure out what ' +
+                                    'happened.\n');
+                        console.log('  Your error number is: ' + err.errno);
+                        console.log('\n  Exiting...\n');
+                    }
+                })
+            // Exits the process after storing user info in the db
+            .then(
+                function() {
```
iankronquist: Functions which take no arguments and just call another function are not necessary. Delete lines 78 and 80.

tschuy: Yes they are. They're only useless if you don't need to pass anything to the function they wrap.

In this instance, try calling ``process.exit()`` with no params. If that works, **then** lines 78 and 80 are unnecessary.

iankronquist: This works just fine for me: `.then( process.exit(0)); `

athai: Deleting lines 78 and 80 makes the error catching fail. It also doesn't store the newly created account into the db correctly. Ditto for line 43.

tschuy: I think you should probably just leave them.

tschuy:
I don't think those comments are necessary.

athai:
The ones about the functions and why they're there?

tschuy:
Yeah. That's just standard fare for a promise, there's no extra explanation (other than a good blog post explaining promises) needed.

ElijahCaine:
@athai Everything here looks good. Your examples in the docs are very good and the behavior described is accurate to experience. It gets the job done.

We may need to add logic like ensuring the user has entered a secure password (we can create a user with password ``1`` right now) but that can be decided on later as we have to make design choices before that code can be written.

tldr: Looks good to me. +1

tschuy:
Looks good to me too :) +1

Kennric:
+1

PR #77, opened 2015-07-30
Add latte command

tschuy:
It runs the linter and then the tests, before printing test coverage.

Is there a better order?

### DIFF
--- a/scripts/latte.sh
+++ b/scripts/latte.sh
@@ -1,0 +1,3 @@
+npm run linter
+npm run coverage
+echo "See coverage/index.html for full coverage information."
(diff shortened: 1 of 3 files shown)


scripts/latte.sh
```diff
@@ -0,0 +1,3 @@
+npm run linter
+npm run coverage
+echo "See coverage/index.html for full coverage information."
```
iankronquist: Could you make this a link like: `file:///path/to/file/coverage/index.html`? I think it might be clickable then, but I'd have to check.

tschuy:
@osuosl/devs

scripts/latte.sh
tschuy: Riiight, I'm in full sh now, so I have lots of space to do things like get `pwd`.

Kennric:
Pretty simple, I see no issues with it. +1

PR #78, opened 2015-07-30
Add Grunt to automatically restart the server

tschuy:
This can be really annoying -- you're debugging an endpoint with your favorite REST client, and you have to restart the server every. Time. Something. Is. Changed.

This adds Grunt, "the Javascript Task Runner", so that we can just run Grunt without running into restart issues.

Docs in /docs. tl;dr:

```
$ npm install
$ npm run grunt
```

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,0 +1,25 @@
+module.exports = function(grunt) {
+
+  grunt.initConfig({
+    watch: {
+      js: {
+        files: [
+          'src/*.js'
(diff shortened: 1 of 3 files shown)


iankronquist:
http://blog.keithcirkel.co.uk/why-we-should-stop-using-grunt/
Thoughts?

tschuy:
You have a good point. Opening a new PR for adding nodemon instead.

PR #79, opened 2015-07-31
Elijah/datetime helper

ElijahCaine:
Note: I branched off of the tschuy/minor_test_refactor. Once that gets merged this should look more reasonable. I will also do a rebase after that happens.

Once the changes from  tschuy/minor_test_refactor are merged there will be less commits. As it stands the only changes happened in ``src/helpers.js`` and ``tests/helpers.js``.

### DIFF
--- a/tests/helpers.js
+++ b/tests/helpers.js
@@ -6,22 +6,22 @@ module.exports = function(expect) {
     describe('checkDateTime helper function', function() {
         it('Returns Unix Timestamp for valid date', function() {
             expect(helper.checkDateTime('2015-04-28')
-                         .to.equal(1430179200000));
+                         .to.equal(Date.parse('2015-04-28'));
         });
 
         it('Returns Unix Timestamp for valid dateTime', function() {
             expect(helper.checkDateTime('2015-04-28 16:30')
-                         .to.equal(1430264400000));
+                         .to.equal(Date.parse('2015-04-28 14:30'));
         });
 
         it('Returns `false` for a non-valid date', function() {
             expect(helper.checkDateTime('2015-04-82')
-                         .to.equal(false));
+                         .to.equal(undefined));
         });
 
         it('Returns `false` for a non-valid dateTime', function() {
             expect(helper.checkDateTime('2015-04-28 24:40')
-                         .to.equal(false));
+                         .to.equal(undefined));
         });
     });
 };
--- a/tests/helpers.js
+++ b/tests/helpers.js
@@ -3,7 +3,25 @@
 var helper = require('../src/helpers');
 
 module.exports = function(expect) {
-    // This line exists to please the linters and tests
-    // Remove it when you write real tests.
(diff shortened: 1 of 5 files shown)


tests/helpers.js
```diff
@@ -0,0 +1,27 @@
...
+        });
+
+        it('Returns Unix Timestamp for valid dateTime', function() {
+            expect(helper.checkDateTime('2015-04-28 16:30')
+                         .to.equal(1430264400000));
+        });
+
+        it('Returns `false` for a non-valid date', function() {
+            expect(helper.checkDateTime('2015-04-82')
+                         .to.equal(false));
```
tschuy: I think undefined might be a better return? It's still falsey so we can still treat it essentially like false.

tests/helpers.js
```diff
@@ -0,0 +1,27 @@
...
+module.exports = function(expect) {
+    describe('checkDateTime helper function', function() {
+        it('Returns Unix Timestamp for valid date', function() {
+            expect(helper.checkDateTime('2015-04-28')
+                         .to.equal(1430179200000));
+        });
+
+        it('Returns Unix Timestamp for valid dateTime', function() {
+            expect(helper.checkDateTime('2015-04-28 16:30')
+                         .to.equal(1430264400000));
```
tschuy: 1430263800000 ?

tests/helpers.js
```diff
@@ -0,0 +1,27 @@
+// test/helpers.js
+
+var helper = require('../src/helpers');
+
+module.exports = function(expect) {
+    describe('checkDateTime helper function', function() {
+        it('Returns Unix Timestamp for valid date', function() {
+            expect(helper.checkDateTime('2015-04-28')
+                         .to.equal(1430179200000));
```
tschuy: 1430179200 ?

iankronquist:
Maybe I don't quite understand. Are we reimplementing [Date.parse](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse)? It parses a string date in one of several formats and returns either a Date object or undefined. Is that what checkDateTime is supposed to do?

tschuy:
We're writing a wrapper for it. I'm ... not entirely sure why.

iankronquist:
Maybe your time is better spent elsewhere? Luckily most of this can be reused in other helper functions.

ElijahCaine:
@tschuy @iankronquist Do you think just using Date.parse instead of wrapping it is better? This was just on the todo list. The actual function checkDateTime is a wrapper for Date.parse.

iankronquist:
I think we should just use Date.parse. It does what we need. We don't need to spend time testing, reviewing, and maintaining it.

tschuy:
yeah, personally I'd be in favor of scrapping this

ElijahCaine:
functionality is not necessary. Closing PR

PR #80, opened 2015-07-31
lint scripts

tschuy:
Minor oversight.


iankronquist:
+1

ElijahCaine:
I broke something in the scripts, it caught that and complained very loudly. Looks good  to me. +1

PR #83, opened 2015-07-31
add nodemon

tschuy:
nodemon will watch the directory, and restart node every time a file
is saved. This means no more forgetting to restart your server after
making changes!

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,12 @@ To start a local instance running on port 8000, just run:
 
 ```
 $ npm install
-$ npm start
+$ npm run devel
 ```
+
+``npm run devel`` is a convenience that automatically restart the server
+every time source files are changed. The standard ``npm start`` still works,
+and will not restart the server automatically.
 
 To run the test suite and linter run:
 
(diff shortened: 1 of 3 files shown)


iankronquist:
I like this solution much better.

iankronquist:
So, I think we will want to make a separate `npm devel` command. We probably won't want to use nodemon in production. Also, nodemon is a dev tool and should be put in devDependencies. Actually all the testing tools should.

tschuy:
For running in production we can run ``node ./src/app.js``, or at least that's what I was thinking.

iankronquist:
I was thinking we would run `npm start` in production. Thoughts?

tschuy:
Your way makes sense, I'll change it to that.

Kennric:
I am happy with this, but a consensus among devs who will be using it would be good.

+1

iankronquist:
Should this be mentioned in the README too?

iankronquist:
I like what I see and this seems pretty handy. +1.

Kennric:
I like the devel command idea, and the readme update is definitely helpful. +1

PR #85, opened 2015-07-31
checkActivities helper function

tschuy:


tschuy:
the tl;dr is that it will throw with a list of all of the failing slugs, or continue cleanly with a list of activity ids.

ElijahCaine:
@tschuy +1 on the tests. We had a small meatspace conversation about them, I think they test enough use cases (and failure cases) to move forward.

tests/helpers.js
```diff
@@ -1,9 +1,36 @@
...
+var helpers = require('../src/helpers');
 
 module.exports = function(expect) {
-    // This line exists to please the linters and tests
-    // Remove it when you write real tests.
-    expect(helper);
+    describe('checkActivities', function() {
+        it('returns a list of activities IDs for proper slugs', function(done) {
+            helpers.checkActivities(['docs', 'dev']).then(function(activities) {
+                expect(activities).to.deep.equal([1, 2]);
```
iankronquist: Silly question, if the test Promise is rejected and the rejection is not caught, does the test fail?

tschuy: Yes, since literally what Mocha does is catch errors.

tests/helpers.js
```diff
@@ -1,9 +1,37 @@
...
+        it('throws when passed a list containing bad slugs', function(done) {
+            helpers.checkActivities(['docs', 'dev', 'cats', 'dogs']).then()
+            .catch(function(err) {
+                err.sort();
+                expect(err).to.deep.have.same.members(['cats', 'dogs'].sort());
+                done();
+            });
+        });
+
+        it('throws when passed a list containing a null slug', function(done) {
```
tschuy: change this to not mention list and rebase

src/helpers.js
```diff
@@ -86,3 +86,40 @@ module.exports = function(app) {
 
     return helpers;
 };
+// src/helpers.js
+
+var app = require('./app');
+var knex = app.get('knex');
+//var errors = require('./errors');
+
+module.exports = {
```
LyonesGamer: I think we might have had some sort of merge error here or something. Why isn't checkActivities() a part of helpers like the other functions in the file?

tests/helpers.js
```diff
@@ -135,3 +135,40 @@ module.exports = function(expect, app) {
...
+        it('returns a list of activities IDs for proper slugs', function(done) {
+            helpers.checkActivities(['docs', 'dev']).then(function(activities) {
+                expect(activities).to.deep.have.same.members([1, 2]);
+                done();
+            });
+        });
+
+        it('throws when passed undefined', function(done) {
+            helpers.checkActivities(undefined).then().catch(function(err) {
+                expect(err).to.be.an('undefined');
```
tschuy: We should change the return value to be explicit, like we did with checkProject: https://github.com/osuosl/timesync-node/blob/develop/tests/helpers.js#L117

LyonesGamer:
@tschuy I made that fix and also merged the objects since checkActivities was kind of hanging out on its own. Does it look good?

Kennric:
Looks good, nice and simple. +1

PR #86, opened 2015-07-31
Add auth errors, auth error tests

tschuy:
Super simple.

### DIFF
--- a/tests/errors.js
+++ b/tests/errors.js
@@ -1,0 +1,23 @@
+var errors = require('../src/errors');
+
+module.exports = function(expect) {
+    describe('6: errorInvalidUsername', function() {
+        it('returns invalid username error', function(done) {
+            err = errors.errorInvalidUsername('bob');
+            expect(err.status).to.equal(401);
+            expect(err.error).to.equal('Invalid username');
+            expect(err.text).to.equal('bob is not a valid username');
+            done();
+        });
+    });
+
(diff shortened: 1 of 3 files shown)


tests/errors.js
```diff
@@ -0,0 +1,23 @@
...
+        });
+    });
+
+    describe('7: errorAuthenticationFailure', function() {
+        it('returns authentication failure block', function(done) {
+            err = errors.errorAuthenticationFailure('Invalid key');
+            expect(err.status).to.equal(401);
+            expect(err.error).to.equal('Authentication failure');
+            expect(err.text).to.equal('Invalid key');
+            done();
```
iankronquist: This seems excessive to me. If we look at the logic which is being tested, we're really just testing the createError function.

ElijahCaine:
@tschuy Tests pass, everything seems pretty solid here.
As far as @iankronquist's comment, I think adding designated createError tests would be beneficial, however as it stands this gets the job done. createError tests ought to be a separate pull request so I'm gonna give this a +1.

ElijahCaine:
@tschuy Tests pass, everything seems pretty solid here.
As far as @iankronquist's comment, I think adding designated createError tests would be beneficial, however as it stands this gets the job done. createError tests ought to be a separate pull request so I'm gonna give this a +1.

iankronquist:
Looks good +1.

PR #87, opened 2015-07-31
checkProject helper function

tschuy:


tschuy:
tl;dr of those tests: if then is called, the project was good and the result is the project id.

If catch is called, the project was bad, and the exact value passed to the function is returned.

ElijahCaine:
These tests are complaint with the API and don't pass (tests that don't pass: it's a feature). +1 from me.

iankronquist:
So question. This PR is not going to be merged as is because it would break Travis. Should we `+1` PRs which is not ready to be merged??

ElijahCaine:
@iankronquist We don't merge these PRs. I guess we need different verbage. Essentially the workflow is this:

1. Create the issue
2. Assign the issue to the person writing the test
3. Person writes test
4. Person submits PR
5. Tests get reviewed and approved.
6. PR gets assigned to person who will write the code to pass the tests.
7. Once the code passes the approved tests the code gets merged.

The +1 in this case means the tests are good and the implementation can be written now.

iankronquist:
It's just a little odd to see the terminology overloaded that way. I could see glancing at a PR, seeing an old test related +1 and a new implementation related +1, thinking "1 + 1 = 2", and hitting the big green button.

ElijahCaine:
Umm.... if you see a +1 and then a commit your counter should be reset to 0. We (almost/should) never merge without +2 after the last commit.

tests/helpers.js
```diff
@@ -1,9 +1,34 @@
...
+var helpers = require('../src/helpers');
 
 module.exports = function(expect) {
-    // This line exists to please the linters and tests
-    // Remove it when you write real tests.
-    expect(helper);
+    describe('checkProject', function() {
+        it('returns a project ID for proper slug', function(done) {
+            helpers.checkProject('ganeti-webmgr').then(function(project) {
+                expect(project).to.equal(1);
```
Kennric: What if I call the API with a slug = '1' or 1? (or any actual project id)?

tschuy: Then it fails. The API only accepts slugs.

Kennric: Where are we testing that it only accepts slugs and that everything else fails?

tschuy: That'd be essentially part of the "throws when passed a bad slug" test. I could add a case explicitly for integers, but I don't think that would really add anything beneficial. We're checking against the slugs, and integers generally aren't slugs. (Though they could be.)

What would the benefit in testing it doesn't accept the "anything else" that isn't gotten from testing a bad slug?

Kennric: You aren't testing bad slugs, you are testing nonexistant slugs. Where do you test that the slug isn't garbage? Or javascript? Or something malicious designed to highjack your query? 

What I am saying specifically about integers: you accidentally send 1 as a slug, the method fails, and sends you back the failed value, 1. Which looks like it sent you back a project id. Confusing, possibly, a better response would be "hey, that's not even a slug".

tschuy:
Blocked by #103 so that we can get some legit slug checking in here.

tschuy:
Now we have the problem that we need to distinguish the kind of check failure that occurred. Should the checkX functions return an object with a ``type:`` and a ``value:``, so for instance ``{type: 'invalid', value: '$@puppies'}`` and ``{type: 'nonexistent', value: 'puppies'}``?

LyonesGamer:
I like that idea, 1) because it clarifies what the error is, and 2) because it clarifies that return values are in fact errors. I feel like code like

```javascript
helpers.checkProject(undefined).then().catch(function(err) {
    // this means that the time.slug was undefined
    expect(err).to.equal(undefined);
    done();
});```

looks ugly. If the existence of a return value indicates an error, then errors should be truthy. Returning undefined or null and expecting it to still mean an error is just wrong, frankly.

File: src/helpers.js [hunk 2]
```diff
@@ -53,5 +53,30 @@ module.exports = {
...
+            knex('projectslugs').select('project')
+            .where('name', slug).then(function(project) {
+                if (project.length === 0) {
+                    // project doesn't exist -- it could be null, undefined,
+                    // invalid, etc.
+                    reject({type: 'nonexistent', value: slug});
+                } else {
+                    resolve(project[0].project);
+                }
+            }).catch(function(err) { reject(err); });
```
LyonesGamer: Should try to be consistent. Probably `reject({type: 'server', value: err})`

LyonesGamer:
Looks good, and tests and linter pass. +1

Kennric:
This looks pretty good to me too, I also like returning an object and type from check methods (do we need to retrofit others?) +1

tschuy:
We will need to retrofit, and I'll open another issue for that.

PR #89, opened 2015-08-03
Thai/check owner eq user

athai:
Checks that the user logged in is the same as the user listed in the JSON object. A test checking whether the user making the post has admin. rights will need to be added later.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -5,16 +5,14 @@ var app = require('./app');
 var errors = require('./errors');
 
 module.exports = {
-    checkUser: function(username, objUser)
-        .then(function(){
-            if (username == objUser) {
-                return true;
+    checkUser: function(username, objUser) {
+        return new Promise(function(resolve, reject) {
+            if (username === objUser) {
+                return resolve(userID);
+            }else {
+                return reject();        
             }
-        })
-        .catch(function(){
-            if (username !== objUser) {
-                return false;
-            }
-        })
+        }
+    }
 };
 
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -5,14 +5,16 @@ var app = require('./app');
 var errors = require('./errors');
 
 module.exports = {
-    checkUser: function(username, objUser) {
-        if (username == objUser) {
-            return true;
-        }
-        if (username !== objUser) {
-            console.log('An error has occurred');
-            return false;
-        }
-    }
+    checkUser: function(username, objUser)
+        .then(function(){
+            if (username == objUser) {
+                return true;
+            }
+        })
+        .catch(function(){
+            if (username !== objUser) {
+                return false;
+            }
+        })
 };
 
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -1,7 +1,18 @@
 // src/helpers.js
 
+var app = require('./app');
+//var knex = app.get('knex');
+var errors = require('./errors');
+
 module.exports = {
-    checkProject: function(username) {
-        username = null;
+    checkUser: function(username, objUser) {
+        if (username == objUser) {
+            return true;
+        }
+        if (username !== objUser) {
+            console.log('An error has occurred');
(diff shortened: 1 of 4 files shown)


tschuy:
Looks good to me! 

@ElijahCaine can we get another +1 or some feedback?

src/helpers.js
```diff
@@ -1,5 +1,23 @@
...
-module.exports = {
+var app = require('./app');
+var knex = app.get('knex');
 
+module.exports = {
+    checkUser: function(username, objUser) {
+        return new Promise(function(resolve, reject) {
+            if (username === objUser) {
+                knex('users').select('id')
+                .where('username', username).then(function(users) {
```
tschuy: where 'username' is objUser, though this won't come up until we decide on using active users or not.

src/helpers.js
```diff
@@ -1,5 +1,23 @@
 // src/helpers.js
 
-module.exports = {
+var app = require('./app');
```
iankronquist: Won't this start up a server when you import it?
I think you'll want to use the export a function pattern: https://github.com/osuosl/timesync-node/blob/develop/src/times.js#L1

src/helpers.js
```diff
@@ -1,5 +1,23 @@
...
+var knex = app.get('knex');
 
+module.exports = {
+    checkUser: function(username, authUser) {
+        return new Promise(function(resolve, reject) {
+            if (username === authUser) {
+                knex('users').select('id')
+                .where('username', username).then(function(users) {
+                    /* Knex only returns lists and not single items. 'users'
```
iankronquist: You want this: http://knexjs.org/#Builder-first

tschuy: TIL that's a thing. Huh.

src/helpers.js
tschuy: I mean, I've not seen any extra "app now listening on"s pop up when we do this, and this isn' t the first place we've done it.

iankronquist: I know that some testing frameworks swallow stdout so that's not exactly proof of anything.

tschuy: mocha does not. I use console.log development with mocha alllll the time.

iankronquist: Ah, you're now exporting the app in `app.js`. I think that changes things. In any case this differs from what we do elsewhere and I think we should be consistent.

athai: How should this be changed then?

tschuy: I don't believe it has to be chagned.

iankronquist: I think either this should be changed to use the `function(app)` pattern as linked above or we should abandon that for this approach. It's not strictly necessary but it looks fairly easy to pull off.

iankronquist:
`@athai 	The linter cares too much` ~ That is an excellent commit message.

tschuy:
This looks good to me! nice job :)

tests/helpers.js
```diff
@@ -1,9 +1,22 @@
...
-    // Remove it when you write real tests.
-    expect(helper);
+    describe('checkUser', function() {
+        it('Returns user ID if username == user', function(done) {
+            helpers.checkUser('tschuy', 'tschuy').then(function(userID) {
+                expect(userID).to.equal(2);
+                done();
+            });
+        });
+        // Later: Include a test that checks if username is admin
```
ElijahCaine: Should we add this test now or is it a feature that will be implemented later (along with corresponding tests)?

athai: Admin rights is a feature that will be implemented later. The comment is just acting as a placeholder/reminder.

ElijahCaine:
Beyond that one nit-pick about adding the admin test, this gets a +1 from me.

ElijahCaine:
Great. Looks good to me.

**official :+1:**

tschuy:
This looks good to me :)

Issue #59, opened 2015-07-27
[8] Post /times endpoint


ElijahCaine:
time should be non-negative
user making submission has valid authentication to make the request

tschuy:
This isn't a blocker, it's **blocked**.

LyonesGamer:
Description:
Use complete endpoint tests to wrap together tests for existing check functions and for actual data entry and then implement data entry (which calls check functions first).

PR #90, opened 2015-08-04
Allow POST /times to create new time entries

LyonesGamer:
Endpoint tests which wrap the existing tests for specific helper functions, and then implementation of actual POST function. Implementation of issue #59.


LyonesGamer:
Rebased to newest develop commit.

tests/times.js
```diff
@@ -90,4 +90,430 @@ module.exports = function(expect, request, baseUrl) {
...
+            });
+        });
+
+        it('should fail with a bad activity', function() {
+            var time = {
+                //jscs:disable
+                duration: 20,
+                user: 'tschuy',
+                project: 'pgd',
+                activities: ['dev', 'docs', 'dancing'],
```
tschuy: Relevant song:

[IMAGE ALT TEXT HERE](http://www.youtube.com/watch?v=Vhf5cuXiLTA)

tests/times.js
```diff
@@ -90,4 +90,430 @@ module.exports = function(expect, request, baseUrl) {
...
+                user: 'tschuy',
+                project: 'project? we need a project?',
+                activities: ['dev', 'docs'],
+                notes: '',
+                issue_uri: 'https://github.com/osuosl/pgd/issues/1',
+                date_worked: '2015-07-30'
+                //jscs:enable
+            };
+
+            var postArg = {
```
tschuy: Can you move this to the describe so that we don't have it repeated twenty times?

tests/times.js
```diff
@@ -90,4 +90,430 @@ module.exports = function(expect, request, baseUrl) {
...
+            };
+
+            request.post(baseUrl + 'times/', postArg, function(err, res) {
+
+                expect(err).to.be(null);
+                expect(res.statusCode).to.be(200);
+
+                createdAt = Date.now().getTime() / 1000;
+                request.get(baseUrl + 'times', function(err, res, body) {
+                    var expectedResults = [
```
LyonesGamer: It does not.  POST returns an empty 200 on success. That's why I do a GET to verify results.

tests/times.js
```diff
@@ -90,4 +90,430 @@ module.exports = function(expect, request, baseUrl) {
...
+            var time = {
+                //jscs:disable
+                duration: -20,
+                user: 'tschuy',
+                project: 'pgd',
+                activities: ['dev', 'docs'],
+                notes: '',
+                issue_uri: 'https://github.com/osuosl/pgd/issues/1',
+                date_worked: '2015-07-30'
+                //jscs:enable
```
LyonesGamer: 1) we were already doing that on the tests that already existed, 
2) it keeps jscs from complaining about having identifiers that aren't camelCase .

tests/times.js
LyonesGamer: Move which? The Auth decorator, or the time object itself?

tests/times.js
```diff
@@ -90,4 +90,430 @@ module.exports = function(expect, request, baseUrl) {
             });
         });
     });
+
+    describe('POST /times', function() {
+        it('should create a new time', function(done) {
```
LyonesGamer: Willingly, although I would point out that all the tests in here use should -- and have since before I got here. If you want to change them all, though, I'm perfectly okay with that, I do like it better without should. Just, we'd have to change them all.

tests/times.js
tschuy: well, both I guess. I was just thinking of the auth postArg object, but the time entry could be pulled out and then explicitly modified each time. Should make sure that that won't run into reference issues though, I don't know if Mocha reloads the entire describe() each time and my guess is it doesn't.

But definitely pull out the postArg.

tests/times.js
tschuy: We were doing that on the tests that had super long passwords in them or directly around snake_case items, I thought.

I'd prefer to keep jscs disabled in a minimum of places -- can you just wrap this around the ones with _?

tests/times.js
```diff
@@ -90,4 +90,430 @@ module.exports = function(expect, request, baseUrl) {
...
+                user: 'tschuy',
+                project: 'pgd',
+                activities: ['dev', 'docs'],
+                notes: '',
+                issue_uri: 'https://github.com/osuosl/pgd/issues/1',
+                date_worked: '2015-07-30'
+                //jscs:enable
+            };
+
+            var postArg = {
```
LyonesGamer: I didn't necessarily figure it could. But the endpoint still can. If it does something like:

if (checkAuthentication(body.auth.user, body.auth.password)) {...} 

var object = body.object;

//process object

tests/times.js
tschuy: Oops, my bad. Misread the test a little.

tests/times.js
tschuy: It does, and it's a minor change. i'll get tests and a PR ready.

tests/times.js
tschuy: all of the new ones we're adding don't have should. Check out tests/login.js, any of the check* function tests that aren't merged yet, etc.

We can fix other tests in the future, when we also theoretically move from 4-2 spaces.

tschuy: radens has a link somewhere explaining why.

tests/times.js
tschuy: Actually, can you test the POST resposne body as well then? If you say it should be empty, it should be empty.

LyonesGamer: I don't follow.  You want the post to return the ID of the object?

tests/times.js
```diff
@@ -90,4 +90,430 @@ module.exports = function(expect, request, baseUrl) {
...
+                                '/whats-fresh-api/issues/56',
+                            date_worked: null,
+                            created_at: null,
+                            updated_at: null,
+                            id: 1
+                        },
+                        {
+                            duration: 20,
+                            user: 'tschuy',
+                            project: ['pgd'],
```
LyonesGamer: Project is a list because it's a list of all slugs for that project. Or at least I'm pretty sure it always has been.

tests/times.js
LyonesGamer: Makes sense. Simple change to make.

tests/times.js
LyonesGamer: I can do that.

tests/times.js
tschuy: sorry, temporarily thought we'd changed this out for a single one instead of activities :)

tests/times.js
```diff
@@ -90,4 +90,430 @@ module.exports = function(expect, request, baseUrl) {
...
+                            date_worked: '2015-07-30',
+                            created_at: createdAt,
+                            updated_at: createdAt,
+                            id: 2
+                            //jscs:enable
+                        }
+                    ];
+                    expect(err).to.be(null);
+                    expect(res.statusCode).to.be(200);
+                    expect(JSON.parse(body)).to.eql(expectedResults);
```
tschuy: ``expect().to.deep.have.same.members()`` would be more optimal here.

tests/times.js
```diff
@@ -90,4 +90,430 @@ module.exports = function(expect, request, baseUrl) {
...
+                    error: 'Invalid foreign key',
+                    status: 409,
+                    text: 'The time does not contain a valid activity reference'
+                };
+
+                expect(jsonBody).to.eql(expectedResult);
+                expect(res.statusCode).to.equal(400);
+
+                request.get(baseUrl + 'times', function(err, res, body) {
+                    var expectedResults = [
```
tschuy: can expectedResults for the failure cases also be factored into the describe() block? You can also then just do
```
                    var expectedResults = initialProjects.concat([
                          {
                              uri: 'https://github.com/osuosl/timesync-node'
                              owner: 'tschuy',
                              slugs: ['ts', 'timesync'],
                              name: 'TimeSync Node'
                          }
                    ])
```

(copied from tests I'm writing)

for your success state, which to me seems like it adds a level of detail about what the test is doing -- it's the initial state, plus the new one.

tschuy:
One more thing! These ``request.post`` bits are all encoding as form data, I think. We'll need to make sure to encode as json.

tschuy:
I've been working on the cleanest way to express these tests. Here's what I have so far: https://github.com/osuosl/timesync-node/pull/97/files#diff-24afa6a7d620f09ab56bf8894916dc61R158

src/errors.js
```diff
@@ -92,7 +92,7 @@ module.exports = {
     errorBadObjectInvalidField: function(objectType, fieldName, expectedType,
     receivedType) {
         return createError(400, 'Bad object', 'Field ' + fieldName + ' of ' +
-            objectType + ' should be a ' + expectedType +
+            objectType + ' should be ' + expectedType +
```
tschuy: If you change the error spec submit a PR for that too :P

LyonesGamer: Actually a typo correction to fit the spec.

tests/times.js
```diff
@@ -83,11 +84,312 @@ module.exports = function(expect, request, baseUrl) {
...
+            request.post(postArg, function(err, res) {
+                var jsonBody = JSON.parse(body);
+                var expectedResult = {
+                    error: 'Invalid foreign key',
+                    status: 409,
+                    text: 'The time does not contain a valid activity reference'
+                };
+
+                expect(jsonBody).to.deep.equal(expectedResult);
+                expect(res.statusCode).to.equal(400);
```
tschuy: 409 or 400?

tschuy:
Other than that, looks good to me. :+1:

tschuy:
Yep, +1

ElijahCaine:
I don't think this has anything to do with these tests specifically, but not all of the tests are being run because of the ``"after each" hook`` which spits out ``Uncaught SyntaxError: Unexpected token C``. 

LINK: https://pastebin.osuosl.org/30481/

LyonesGamer:
Blocked waiting for #85

LyonesGamer:
With the checkActivities() function from PR #85, tests all pass. I consider this feature complete, simply blocked waiting for #85.

src/times.js
```diff
@@ -34,6 +38,7 @@ module.exports = function(app) {
                     return res.send(times);
                 }
             }).catch(function(error) {
+                console.log('Get users failed.');
```
tschuy: All the comments!

tests/times.js
```diff
@@ -84,11 +85,308 @@ module.exports = function(expect, request, baseUrl) {
...
+    describe('POST /times', function() {
+        function getPostObject(uri, time) {
+            return {
+                uri: uri,
+                json: true,
+                body: {
+                    auth: {
+                        type: 'password',
+                        user: 'tschuy',
+                        password: '$2a$10$6jHQo4XTceYyQ/SzgtdhleQqkuy2G27omuI' +
```
tschuy: add an authentication test ;)

tests/times.js
```diff
@@ -84,11 +85,308 @@ module.exports = function(expect, request, baseUrl) {
...
+                request.get(baseUrl + 'times', function(err, res, body) {
+                    expect(err).to.equal(null);
+                    expect(res.statusCode).to.equal(200);
+                    expect(JSON.parse(body)).to.deep.equal(initialData);
+                    done();
+                });
+            });
+        });
+
+        it('fails with a bad owner', function(done) {
```
tschuy: You also need to test for invalid datatypes. What happens if I send owner as an array or int? Yes, we put that in the error spec.

tests/times.js
```diff
@@ -84,11 +85,818 @@ module.exports = function(expect, request, baseUrl) {
...
+                    object: time
+                }
+            };
+        }
+
+        var initialData = [
+            {
+                duration: 12,
+                user: 'tschuy',
+                project: ['wf'],
```
ElijahCaine: is this a string or an array? ``initialData`` makes it an array, later you set project to a string (line 134 for instance)

LyonesGamer: For the argument passed into a POST call, `project` is a string which represents *one* slug of the associated project. For the data returned from a GET call (which is what `initialData` is for), `project` is an array which represents *every* slug of the associated project. Admittedly it is a little confusing.

ElijahCaine:
Tests pass, that is promising.
I was only able to get through about 1/2 of this. Nothing jumped out at me except for the one comment I left. +0.5

LyonesGamer:
Rebased to develop

src/times.js
```diff
@@ -206,4 +211,182 @@ module.exports = function(app) {
...
+            }
+
+            if (helpers.getType(time.user) !== 'string') {
+                let err = errors.errorBadObjectInvalidField('time', 'user',
+                    'username', helpers.getType(time.user));
+                return res.status(err.status).send(err);
+            }
+
+            //Test issue URI
+            //jscs:disable
```
Kennric: Can we just remove these lines? Or is there a reason to keep them in, but commented?

LyonesGamer: The jscs disable/enable lines? The syntax is that they're comments. They aren't valid JavaScript. They're basically pragmas.

tschuy: Just merge https://github.com/osuosl/timesync-node/pull/119 and then we can
get rid of them :)

Kennric: Yeah, I wasn't sure if that was ready yet, but I'd like to see that merged and this rebased and these removed.

Kennric:
Aside from jscs lines, this looks pretty good. I am going to +1, but I'd like to see two more, just because it is so large.

src/times.js
```diff
@@ -200,10 +200,178 @@ module.exports = function(app) {
...
+            }
+
+            if (time.duration < 0) {
+                let err = errors.errorBadObjectInvalidField('time', 'duration',
+                    'positive number', 'negative number');
+                return res.status(err.status).send(err);
+            }
+
+            //Test existence and type of project
+            if (!time.project) {
```
tschuy: Using validateFields might make this 100-line section less ungodly large https://github.com/osuosl/timesync-node/blob/develop/src/helpers.js#L61

src/times.js
```diff
@@ -200,10 +200,133 @@ module.exports = function(app) {
...
+
+            // Test duration value
+            if (time.duration < 0) {
+                let err = errors.errorBadObjectInvalidField('time', 'duration',
+                    'positive number', 'negative number');
+                return res.status(err.status).send(err);
+            }
+
+            //Test each activity
+            for (let activity of time.activities) {
```
tschuy: you'll also want to test for the strings being bad slugs. (Same for project.)

tschuy:
After you fix anything @Kennric points out, this gets a +1 from me :)

Kennric:
I don't see any red flags. There's a lot of material, so something could slip through, but we'll fix it when it does. +1

PR #92, opened 2015-08-04
Change passport auth block

### LINKED ISSUE #91: [1] Move passport to using auth[*] fields for login tests

tschuy:
Fixes #91

### DIFF
--- a/src/auth/local.js
+++ b/src/auth/local.js
@@ -3,6 +3,10 @@ var bcrypt = require('bcrypt');
 
 module.exports = function(knex) {
     return new LocalStrategy(
+        {
+            usernameField: 'auth[username]',
+            passwordField: 'auth[password]'
+        },
         function(username, password, done) {
             /* done parameters: err, user, information
                authentication succeeds if err is null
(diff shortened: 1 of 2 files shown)


Kennric:
Looks correct, I like the auth {} block structure, I think that will help keep things flexible and extendable. +1

iankronquist:
@tschuy Do you think you should use `knex.first`?

tschuy:
Yeah, I'll get that.

src/auth/local.js
```diff
@@ -3,6 +3,10 @@ var bcrypt = require('bcrypt');
 
 module.exports = function(knex) {
     return new LocalStrategy(
+        {
+            usernameField: 'auth[username]',
+            passwordField: 'auth[password]'
+        },
```
ElijahCaine: @tschuy What do these two lines do with usernameField and passwordField?

tschuy: https://github.com/jaredhanson/passport-local#parameters

ElijahCaine:
Looks good to me. +1

Issue #29, opened 2015-07-10
[8] POST /projects endpoint


tschuy:
To add a new object, POST to */projects/* with a JSON body:

    {
       "uri":"https://code.osuosl.org/projects/ganeti-webmgr",
       "name":"Ganeti Web Manager",
       "slugs":[<slugs>],
       "owner": <username>
    }

tschuy:
Current status: needs #87, which needs #103.

tschuy:
Current status: reviewed, and I'm about to start implementation.

tschuy:
Time spent so far: ???

tschuy:
Time = Time + 5? Maybe?

tschuy:
Time += 2, let's say. Not sure.

PR #97, opened 2015-08-04
POST /projects

tschuy:
Fixes https://github.com/osuosl/timesync-node/issues/29.

These tests are a work in progress -- I have the "success" tests written, and the "it doesn't add things" written, but not the error message tests.

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -4,9 +4,10 @@ function createError(status, name, text, values) {
         error: name,
         text: text
     };
-    if(values) {
+    if (values) {
         err.values = values;
     }
+
     return err;
 }
 
@@ -135,12 +136,13 @@ module.exports = {
        but the object being created uses existing slugs.
     */
     errorSlugsAlreadyExist: function(slugs) {
+        var message;
         if (slugs.length === 1) {
-            var message = 'slug ' + slugs[0] + ' already exists';
+            message = 'slug ' + slugs[0] + ' already exists';
         } else {
-            var message = 'slugs ';
-            for(var i = 0; i < slugs.length; i++) {
-                message = message + slugs[i] + ', '
+            message = 'slugs ';
+            for (var i = 0; i < slugs.length; i++) {
+                message = message + slugs[i] + ', ';
             }
             // chop off last ', '
             message = message.substring(0, message.length - 2);
@@ -148,7 +150,16 @@ module.exports = {
         }
 
         return createError(409, 'The slug provided already exists',
-            message, slugs)
+            message, slugs);
+    },
+
+    /*
+     * Error 9: Authorization failure. Used when a user attempts to do something
+       they aren't allowed to do, but is properly authenticated.
+    */
+    errorAuthorizationFailure: function(user, activity) {
+        return createError(401, 'Authorization failure',
+            user + ' is not authorized to ' + activity);
     },
 
 };
--- a/src/errors.js
+++ b/src/errors.js
@@ -139,12 +139,12 @@ module.exports = {
             var message = 'slug ' + slugs[0] + ' already exists';
         } else {
             var message = 'slugs ';
-            for(slug of slugs) {
-                message.concat(slug + ', ')
+            for(var i = 0; i < slugs.length; i++) {
+                message = message + slugs[i] + ', '
             }
             // chop off last ', '
             message = message.substring(0, message.length - 2);
-            message.concat(' already exist')
+            message = message + ' already exist';
         }
 
         return createError(409, 'The slug provided already exists',
--- a/src/errors.js
+++ b/src/errors.js
@@ -110,10 +110,23 @@ module.exports = {
      * param receivedIdentifier(string): The value that was received from the
      *    client.
      */
-    errorInvalidIdentifier: function(expectedType, receivedIdentifier) {
+    errorInvalidIdentifier: function(expectedType, receivedIdentifiers) {
+        var message;
+        if (!Array.isArray(receivedIdentifiers)) {
+            message = 'Expected ' + expectedType + ' but received ' +
+                receivedIdentifiers;
+            receivedIdentifiers = [receivedIdentifiers];
+        } else {
+            message = 'Expected ' + expectedType + ' but received: ';
+            for (var i = 0; i < receivedIdentifiers.length; i++) {
+                message = message + receivedIdentifiers[i] + ', ';
+            }
+            // chop off last ', '
+            message = message.substring(0, message.length - 2);
+        }
+
         return createError(400, 'The provided identifier was invalid',
-            'Expected ' + expectedType + ' but received ' +
-            receivedIdentifier);
+            message, receivedIdentifiers);
     },
 
     /* Error 6: Authentication failed due to an invalid username. */
--- a/src/errors.js
+++ b/src/errors.js
@@ -1,9 +1,13 @@
-function createError(status, name, text) {
-    return {
+function createError(status, name, text, values) {
+    var err = {
         status: status,
         error: name,
         text: text
     };
+    if(values) {
+        err.values = values;
+    }
+    return err;
 }
 
 module.exports = {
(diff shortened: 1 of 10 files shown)


tschuy:
This contains two proposed errors that I'll be adding to the error documentation.

Additionally, it extends the ``The provided identifier was invalid`` error to support multiple 'but received' items:

``expected slug but receieved $*#*cat and )_!@#mouse``

tests/projects.js
```diff
@@ -99,4 +105,294 @@ module.exports = function(expect, request, baseUrl) {
...
+        var requestOptions = {
+            url: baseUrl + 'projects/',
+            json: true
+        };
+
+        it('successfully creates a new project with slugs', function(done) {
+            requestOptions.form = postArg;
+
+            request.post(requestOptions, function(err, res, body) {
+                console.log(err);
```
Kennric: Debugging statement, left in on purpose?

tschuy:
@Kennric

Kennric:
This appears to test eveything correctly, and include the response bodies as in the newes draft api. +1

tschuy:
@ElijahCaine

ElijahCaine:
Tests seem reasonable. I think we can move on to implementation. +1

tschuy:
All that's left is for #103 and #89 to be merged.

tests/projects.js
```diff
@@ -99,4 +106,294 @@ module.exports = function(expect, request, baseUrl) {
...
+        var requestOptions = {
+            url: baseUrl + 'projects/',
+            json: true,
+            method: 'POST'
+        };
+
+        it('successfully creates a new project with slugs', function(done) {
+            requestOptions.form = postArg;
+
+            request(requestOptions, function(err, res, body) {
```
athai: Shouldn't this be request.post(requestOptions, ...)?

tschuy: It doesn't have to be since we're defining POST in the ``requestOptions`` variable, but for consistency's sake yes.

src/errors.js
```diff
@@ -105,10 +110,23 @@ module.exports = {
...
-    errorInvalidIdentifier: function(expectedType, receivedIdentifier) {
+    errorInvalidIdentifier: function(expectedType, receivedIdentifiers) {
+        var message;
+        if (!Array.isArray(receivedIdentifiers)) {
+            message = 'Expected ' + expectedType + ' but received ' +
+                receivedIdentifiers;
+            receivedIdentifiers = [receivedIdentifiers];
+        } else {
+            message = 'Expected ' + expectedType + ' but received: ';
+            for (var i = 0; i < receivedIdentifiers.length; i++) {
```
LyonesGamer: There's a built-in Javascript function for this: Array.join(separator = ',') https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join

src/errors.js
```diff
@@ -126,4 +144,35 @@ module.exports = {
...
+     * Error 8: Slugs already exist. Used when a new object is being created,
+       but the object being created uses existing slugs.
+    */
+    errorSlugsAlreadyExist: function(slugs) {
+        var message;
+        if (slugs.length === 1) {
+            message = 'slug ' + slugs[0] + ' already exists';
+        } else {
+            message = 'slugs ';
+            for (var i = 0; i < slugs.length; i++) {
```
LyonesGamer: Same here, just Array.join()

src/helpers.js
```diff
@@ -1,57 +1,58 @@
...
+
+              Valid:
+                test
+                a1b2-c3
+                1a2b
+
+              Invalid:
+                a1b2--c3
+            */
+            if (slug === undefined || slug === null) {
```
LyonesGamer: Man I hate Javascript's type coercion. Just wanted to say that, no problems with the code.

src/helpers.js
```diff
@@ -1,57 +1,58 @@
...
+                // Javascript will cast these to "undefined" and "null",
+                // which then pass the test
+                return false;
+            }
 
-        var hasDoubleHyphens = new RegExp('--');
-        var containsLetter = new RegExp('[a-z]+');
-
-        // alphanumeric plus hyphens, but with no external hyphens
-        var alphanumeric = new RegExp('^[a-z0-9]+[a-z0-9-]*[a-z0-9]+$');
```
LyonesGamer: Today I learned that sticking the hyphen at the beginning or end of a class prevents the need for escaping. I think it's still considered good practice to escape hyphens in character classes though.

LyonesGamer: Also, are we specifically disallowing capital letters? Because [a-zA-Z0-9] can be shortened as \w. Alternately, you could consider the slug "test" to be identical to "Test" and just call slug.toLowerCase() on it.

src/helpers.js
```diff
@@ -1,57 +1,58 @@
...
+                return true;
             }
-        });
-    }
+
+            var hasDoubleHyphens = new RegExp('--');
+            var containsLetter = new RegExp('[a-z]+');
+
+            // alphanumeric plus hyphens, but with no external hyphens
+            var alphanumeric = new RegExp('^[a-z0-9]+[a-z0-9-]*[a-z0-9]+$');
```
LyonesGamer: Today I learned that sticking the hyphen at the beginning or end of a class prevents the need for escaping. I think it's still considered good practice to escape hyphens in character classes though.

LyonesGamer: Also, are we specifically disallowing capital letters? Because [a-zA-Z0-9] can be shortened as \w. Alternately, you could consider the slug "test" to be identical to "Test" and just call slug.toLowerCase() on it.

src/errors.js
tschuy: TIL. That makes sense, though. Duh.

src/projects.js
```diff
@@ -145,4 +147,70 @@ module.exports = function(app) {
             return res.status(err.status).send(err);
         });
     });
+
+    app.post(app.get('version') + '/projects', function(req, res) {
+        var err; // used for any error response
+        var obj = req.body.object;
+
+        if (obj.uri && !validUrl.isWebUri(obj.uri)) {
```
LyonesGamer: Wouldn't this let by projects without any uri at all? Or is uri nullable?

src/helpers.js
tschuy: We are in fact. No real reason why.

src/projects.js
tschuy: uri is nullable.

LyonesGamer:
Tests and linter pass, everything else looks good. +1

src/helpers.js
Kennric: One fewer axis for typos. lowercase the lot and we don't have to worry about PgD, gWM, etc

Kennric:
This is looking really good - there is so much in here, though, that I think a third review is a good idea. +1, but please get another before merging.

src/helpers.js
LyonesGamer: That's why I was thinking you make it case insensitive. PgD is the exact same thing as pgd is the same as PGD. But either way really works.

ElijahCaine: As a side note, ``\w`` matches 'word characters' which allegedly includes, at minimum ``[A-Za-z0-9_]`` (notice the underscore) and inconsistently includes other characters like ``'`` as well as a few others depending on implementation.

src/errors.js
```diff
@@ -105,10 +110,19 @@ module.exports = {
...
+    errorInvalidIdentifier: function(expectedType, receivedIdentifiers) {
+        var message;
+        if (!Array.isArray(receivedIdentifiers)) {
+            message = 'Expected ' + expectedType + ' but received ' +
+                receivedIdentifiers;
+            receivedIdentifiers = [receivedIdentifiers];
+        } else {
+            message = 'Expected ' + expectedType + ' but received: ' +
+                receivedIdentifiers.join(', ');
+        }
```
ElijahCaine: is this behavior specified in the api? I don't remember it giving multiple ``Expect ___ but received ___ ,`` messages for array input.

tschuy: https://github.com/osuosl/timesync/pull/11/files#diff-459f373b6fe1c0684fe24c353ae5ee62R127

ElijahCaine:
I receive an error when running ``npm run latte`` ( https://pastebin.osuosl.org/30466/ ). Beyond this the code looks good, the tests pass, and everything in the code makes sense. Once the ``npm run latte`` issue is resolved I can give this a +1

ElijahCaine:
Looks good to me +1

Kennric:
Continues to look good. +1

PR #99, opened 2015-08-05
POST projects/:slug

ElijahCaine:
Updates a given project based on a given slug.

Testing:
```
$ npm test
```
Tests live in ``tests/projects.js`` under ``describe('POST /projects/:slug', ...``

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -81,6 +81,19 @@ module.exports = function(app) {
                     reject({type: 'database', value: err});
                 });
             });
+        },
+
+        getType: function(receivedObject) {
+            // typeof returns object for arrays, so we need a special check
+            if(typeof receivedObject === 'object') {
+                if(Array.isArray(receivedObject)) {
+                    return 'array';
+                } else {
+                    return 'object';
+                }
+            } else {
+                return typeof receivedObject;
+            }
         }
     };
 
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -88,6 +88,8 @@ module.exports = function(app) {
             if (typeof receivedObject === 'object') {
                 if (Array.isArray(receivedObject)) {
                     return 'array';
+                } else if (receivedObject === null) {
+                    return 'null';
                 } else {
                     return 'object';
                 }
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -28,7 +28,7 @@ module.exports = function(app) {
             // if slug is only one character long, shortcut and finish
             // rolling this into the alphanumeric test would conflict with
             // getting rid of beginning/ending hyphens
-            if (RegExp('^[a-z]$').test(slug)) {
+            if (new RegExp('^[a-z]$').test(slug)) {
                 return true;
             }
 
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -77,6 +77,7 @@ module.exports = function(app) {
                         // skip it
                         continue;
                     }
+
                     field.actualType = fieldValue;
                     return field;
                 }
@@ -110,18 +111,18 @@ module.exports = function(app) {
             });
         },
 
-        getType: function(receivedObject) {
+        getType: function(receieved) {
             // typeof returns object for arrays, so we need a special check
-            if (typeof receivedObject === 'object') {
-                if (Array.isArray(receivedObject)) {
+            if (typeof receieved === 'object') {
+                if (Array.isArray(receieved)) {
                     return 'array';
-                } else if (receivedObject === null) {
+                } else if (receieved === null) {
                     return 'null';
                 } else {
                     return 'object';
                 }
             } else {
-                return typeof receivedObject;
+                return typeof receieved;
             }
         }
     };
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -82,6 +82,8 @@ module.exports = function(app) {
                     return field;
                 }
             }
+
+            return null;
         },
 
         checkProject: function(slug) {
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -58,26 +58,27 @@ module.exports = function(app) {
             });
         },
 
-        validateFields: function(object, fields, required) {
-            /* fields is an array of objects with 'type' and 'name' keys.
-               object is the object containing those fields. The value is tested
-               against the type, and if there's a mismatch, validateFields
-               returns an object with the type, name, and actual_type of the
-               field.
+        validateFields: function(object, fields) {
+            /* fields is an array of objects with 'type', 'required' and 'name'
+               keys. object is the object containing those fields. The value is
+               tested against the type, and if there's a mismatch,
+               validateFields returns an object with the type, name, required,
+               and actual_type of the field.
 
-               If required is set to true, the fields are expected to not be
-               undefined. Otherwise, undefined fields will not raise an error.
+               If required is set to true, the field is expected to not be
+               undefined. If false, an undefined field will not raise an error.
             */
 
             for (let field of fields) {
                 let fieldValue = helpers.getType(object[field.name]);
                 if (fieldValue !== field.type) {
-                    if (!required && fieldValue === 'undefined') {
-                        continue; // skip this field if it's allowed to be undef
-                    } else {
-                        field.actualType = fieldValue;
-                        return field;
+                    if (object[field.name] === undefined && !field.required) {
+                        // if the field isn't required, and it's undefined,
+                        // skip it
+                        continue;
                     }
+                    field.actualType = fieldValue;
+                    return field;
                 }
             }
         },
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -1,3 +1,5 @@
+'use strict';
+
 var helpers;
 
 module.exports = function(app) {
@@ -54,6 +56,30 @@ module.exports = function(app) {
                     return reject();
                 }
             });
+        },
+
+        validateFields: function(object, fields, required) {
+            /* fields is an array of objects with 'type' and 'name' keys.
+               object is the object containing those fields. The value is tested
+               against the type, and if there's a mismatch, validateFields
+               returns an object with the type, name, and actual_type of the
+               field.
+
+               If required is set to true, the fields are expected to not be
+               undefined. Otherwise, undefined fields will not raise an error.
+            */
+
+            for (let field of fields) {
+                let fieldValue = helpers.getType(object[field.name]);
+                if (fieldValue !== field.type) {
+                    if (!required && fieldValue === 'undefined') {
+                        continue; // skip this field if it's allowed to be undef
+                    } else {
+                        field.actualType = fieldValue;
+                        return field;
+                    }
+                }
+            }
         },
 
         checkProject: function(slug) {
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -85,8 +85,8 @@ module.exports = function(app) {
 
         getType: function(receivedObject) {
             // typeof returns object for arrays, so we need a special check
-            if(typeof receivedObject === 'object') {
-                if(Array.isArray(receivedObject)) {
+            if (typeof receivedObject === 'object') {
+                if (Array.isArray(receivedObject)) {
                     return 'array';
                 } else {
                     return 'object';
--- a/src/projects.js
+++ b/src/projects.js
@@ -233,9 +233,35 @@ module.exports = function(app) {
     app.post(app.get('version') + '/projects/:slug', function(req, res) {
         var obj = req.body.object;
 
-        // run various checks
+        // valid keys
+        var validKeys = ['name', 'uri', 'owner', 'slugs'];
+        for (let key in obj) {
+            // indexOf returns -1 if the parameter is in
+            // the array, so this returns true if
+            // the slug is not in slugNames
+            if (validKeys.indexOf(key) === -1) {
+                let err = errors.errorBadObjectUnknownField('project', key);
+                return res.status(err.status).send(err);
+            }
+        }
+
+        // check string fields
+        for (let field of ['name', 'uri', 'owner']) {
+          if(obj[field] && helpers.getType(obj[field]) !== 'string') {
+              let err = errors.errorBadObjectInvalidField('project', field, 'string', helpers.getType(obj[field]));
+              return res.status(err.status).send(err);
+          }
+        }
+
+        // check that slugs is array
+        if (obj.slugs && helpers.getType(obj.slugs) !== 'array') {
+            let err = errors.errorBadObjectInvalidField('project', 'slugs', 'array', helpers.getType(obj.slugs));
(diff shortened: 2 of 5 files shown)


tests/projects.js
```diff
@@ -99,4 +99,377 @@ module.exports = function(expect, request, baseUrl) {
...
+
+                body = JSON.parse(body);
+
+                expect(body).to.equal(expectedResults);
+
+                checkListEndpoint(done, expectedResults);
+            });
+        });
+
+        it('doesnt patches a project with bad uri, name, slugs, and and owner',
```
Kennric: non-fatal typos in this line

tests/projects.js
```diff
@@ -99,4 +99,377 @@ module.exports = function(expect, request, baseUrl) {
...
+
+                var expectedResults = JSONcopy(originalProject);
+
+                checkListEndpoint(done, expectedResults);
+            });
+        });
+
+        it('doesnt patch a project with invalid slugs', function(done) {
+            postArg.object = JSONcopy(originalProject);
+            postArg.object.key = badProject.key;
```
Kennric: Shouldn't this be badProject.slugs?

tests/projects.js
```diff
@@ -99,4 +99,377 @@ module.exports = function(expect, request, baseUrl) {
...
+            uri: 'https://code.osuosl.org/projects/'
+        };
+
+        var patchedProjectName  = {name:  patchedProject.name };
+        var patchedProjectOwner = {owner: patchedProject.owner};
+        var patchedProjectUri   = {uri:   patchedProject.uri  };
+        var patchedProjectSlugs = {slugs: patchedProject.slugs};
+
+        var badProject = {
+            name:  ['a name'],
```
LyonesGamer: Any particular reason these are arrays?

ElijahCaine: Arrays are an invalid type for name, owner, and uri, so if we receive them it will throw an invalid type error in the tests.

tests/projects.js
```diff
@@ -107,6 +107,379 @@ module.exports = function(expect, request, baseUrl) {
...
+                          'PvSG8rwN4xyaF5W'
+            },
+        };
+
+        var requestOptions = {
+            url: baseUrl + 'projects/gwm',
+            json: true
+        };
+
+        var checkListEndpoint = function(done, expectedResults) {
```
tschuy: Maybe add a comment saying what this is used for

tests/projects.js
```diff
@@ -107,6 +107,379 @@ module.exports = function(expect, request, baseUrl) {
...
+        it('successfully patches a projects URI, slugs, owner, and name',
+           function(done) {
+            postArg.object = JSONcopy(patchedProject);
+            requestOptions.form = JSONcopy(postArg);
+
+            request.post(requestOptions, function(err, res, body) {
+                expect(err).to.be.a('null');
+                expect(res.statusCode).to.equal(200);
+
+                // Set expected results to the new state of
```
tschuy: new state of ???

tests/projects.js
```diff
@@ -107,6 +107,379 @@ module.exports = function(expect, request, baseUrl) {
...
+                body = JSON.parse(body);
+
+                // expect body of post request to be the new state of gwm
+                expect(body).to.equal(expectedResults);
+
+                checkListEndpoint(done, expectedResults);
+            });
+        });
+
+        it('successfully patches a projects URI', function(done) {
```
tschuy: project's

tests/projects.js
```diff
@@ -107,6 +107,379 @@ module.exports = function(expect, request, baseUrl) {
...
+
+                body = JSON.parse(body);
+
+                expect(body).to.equal(expectedResults);
+
+                checkListEndpoint(done, expectedResults);
+            });
+        });
+
+        it('doesnt patche a project with bad uri, name, slugs, and and owner',
```
tschuy: doesn't patch

tschuy:
10/10, definitely :+1: from me. Now to go copy-paste these and make my own PR for times.

Kennric:
Looks good, minus minor typos already noted.

tschuy:
:+1:

tschuy:
TODO: refactor, fix validateFields

tschuy:
@osuosl/devs This is ready for review :dancer:

src/helpers.js
```diff
@@ -56,6 +58,30 @@ module.exports = function(app) {
...
+               field.
+
+               If required is set to true, the fields are expected to not be
+               undefined. Otherwise, undefined fields will not raise an error.
+            */
+
+            for (let field of fields) {
+                let fieldValue = helpers.getType(object[field.name]);
+                if (fieldValue !== field.type) {
+                    if (!required && fieldValue === 'undefined') {
```
Kennric: Can't you short circuit this before you even load the object? If it is undefined and not required, there is no check to perform, and you can safely skip to the next field.

tschuy: So you're saying something like this:

```
            for (let field of fields) {
                if (object[field.name] === 'undefined' && !required) {
                    continue.
                }
                let fieldValue = helpers.getType(object[field.name]);
                if (fieldValue !== field.type) {
                        field.actualType = fieldValue;
                        return field;
                    }
                }
            }
```

?

src/helpers.js
```diff
@@ -81,6 +107,21 @@ module.exports = function(app) {
                     reject({type: 'database', value: err});
                 });
             });
+        },
+
+        getType: function(receivedObject) {
```
Kennric: This is fine, but the param recievedObject seems like it is assuming this is an object, but then you test to see if it is an object or not. Maybe Object should not be part of the parameter name, since we aren't sure yet and it makes the code a bit more confusing.

tests/projects.js
```diff
@@ -107,6 +107,384 @@ module.exports = function(expect, request, baseUrl) {
...
+            });
+        });
+
+        it("doesn't patch a project with a different owner", function(done) {
+            requestOptions.form = copyJsonObject(postArg);
+            requestOptions.form.auth.user = 'deanj';
+            requestOptions.form.auth.password = 'pass';
+            requestOptions.form.object = copyJsonObject(patchedProject);
+
+            console.log(requestOptions);
```
ElijahCaine: Is there a reason this is still here?

tschuy: that's sure a polite way of saying "don't leave debug code in". :P

src/projects.js
```diff
@@ -229,4 +229,180 @@ module.exports = function(app) {
...
+        let fields = [
+          {name: 'name', type: 'string'},
+          {name: 'uri', type: 'string'},
+          {name: 'owner', type: 'string'},
+          {name: 'slugs', type: 'array'}
+        ];
+
+        // validateFields takes the object to check fields on,
+        // an array of field names and types, and a boolean indicating
+        // whether the fields are required.
```
Kennric: Whether -all- the fields are required? I was expecting fields to individually be required or not, but it looks like there is a required field for them all together. This is a little confusing,

tschuy: We're either patching -- which means we don't need any field in particular -- or we're creating, which means we need all of the fields. 

I could implement granularity, though. Any idea where we would end up using it?

src/projects.js
```diff
@@ -229,4 +229,180 @@ module.exports = function(app) {
...
+        let validationFailure = helpers.validateFields(obj, fields, false);
+        if (validationFailure) {
+            let err = errors.errorBadObjectInvalidField(
+                'project', validationFailure.name, validationFailure.type,
+                validationFailure.actualType);
+            return res.status(err.status).send(err);
+        }
+
+        // check validity of uri
+        if (obj.uri && !validUrl.isWebUri(obj.uri)) {
```
Kennric: what exactly is isWebUri testing? syntax of the uri, or more?

tschuy: yes, just syntax. https://www.npmjs.com/package/valid-url

tschuy: I'll chagne teh comment to make that more clear.

src/helpers.js
```diff
@@ -56,6 +58,30 @@ module.exports = function(app) {
...
+               object is the object containing those fields. The value is tested
+               against the type, and if there's a mismatch, validateFields
+               returns an object with the type, name, and actual_type of the
+               field.
+
+               If required is set to true, the fields are expected to not be
+               undefined. Otherwise, undefined fields will not raise an error.
+            */
+
+            for (let field of fields) {
```
ElijahCaine: Why are there helper functions being added in this PR?

tschuy:
@ElijahCaine because I wrote them to make the POST work properly. They were actually integrated into the POST until I refactored them out.

src/projects.js
```diff
@@ -229,4 +229,180 @@ module.exports = function(app) {
...
+                let err = errors.errorInvalidIdentifier('slug', invalidSlugs);
+                return res.status(err.status).send(err);
+            }
+        }
+
+        obj.slugs = obj.slugs || [];
+
+        // returns the project ID for the project slug
+        let projectIdQuery = knex('projectslugs').select('project')
+        .where('name', req.params.slug);
```
ElijahCaine: I don't remember what we decided on for the style guide, but I believe this should be indented.

tests/helpers.js
```diff
@@ -17,6 +17,78 @@ module.exports = function(expect, app) {
         });
     });
 
+    describe('validateFields', function() {
+        it('returns field when field is missing if required', function(done) {
+            var obj = {string: 'string', array: []};
+            var fields = [
+              {name: 'string', type: 'string', required: true},
+              {name: 'array', type: 'array', required: true},
+              {name: 'integer', type: 'number', required: true},
```
ElijahCaine: These lines should be indented 4 times, not 2.

ElijahCaine:
Tests pass, coverage seems to be all there. There are a few minor style changes that could be made, but it is readable and easy enough to understand. There don't seem to be any logic/code errors just slight stylistic changes that could be made. +1

src/projects.js
tschuy: I'm pretty sure this is what we ended up with. I'll look around for other instances where we've had to break before a period.

tschuy:
we're not indenting things here:

https://github.com/osuosl/timesync-node/blob/develop/src/projects.js#L118

Nor here:

https://github.com/osuosl/timesync-node/blob/develop/src/helpers.js#L50

I think this is the style we ended up landing with.

src/helpers.js
```diff
@@ -56,6 +58,32 @@ module.exports = function(app) {
...
+                    if (object[field.name] === undefined && !field.required) {
+                        // if the field isn't required, and it's undefined,
+                        // skip it
+                        continue;
+                    }
+
+                    field.actualType = fieldValue;
+                    return field;
+                }
+            }
```
LyonesGamer: It may be good to explicitly return false or something here for clarity's sake.

src/projects.js
```diff
@@ -229,4 +229,180 @@ module.exports = function(app) {
...
+          {name: 'name', type: 'string', required: false},
+          {name: 'uri', type: 'string', required: false},
+          {name: 'owner', type: 'string', required: false},
+          {name: 'slugs', type: 'array', required: false}
+        ];
+
+        // validateFields takes the object to check fields on,
+        // an array of field names and types, and a boolean indicating
+        // whether the fields are required.
+        let validationFailure = helpers.validateFields(obj, fields, false);
```
LyonesGamer: Changed the signature of validateFields but didn't get rid of the extraneous argument here. At least it doesn't have any effect.

src/projects.js
```diff
@@ -229,4 +229,180 @@ module.exports = function(app) {
...
+        if (validationFailure) {
+            let err = errors.errorBadObjectInvalidField(
+                'project', validationFailure.name, validationFailure.type,
+                validationFailure.actualType);
+            return res.status(err.status).send(err);
+        }
+
+        // check validity of uri syntax
+        if (obj.uri && !validUrl.isWebUri(obj.uri)) {
+            let err = errors.errorInvalidIdentifier('uri', obj.uri);
```
LyonesGamer: This should be (at least in my head, we may have changed the errors since I designed them) a Bad Object error. Invalid Identifier is for slugs/IDs only, and only for GET requests.

src/projects.js
```diff
@@ -229,4 +229,180 @@ module.exports = function(app) {
...
+        }
+
+        // check validity of slugs
+        if (obj.slugs && obj.slugs.length) {
+            let invalidSlugs = obj.slugs.filter(function(slug) {
+                return !helpers.validateSlug(slug);
+            });
+
+            if (invalidSlugs.length) {
+                let err = errors.errorInvalidIdentifier('slug', invalidSlugs);
```
LyonesGamer: Same here, even those these are slugs, Bad Object should be for POST and Invalid Identifier for GET.

src/projects.js
```diff
@@ -229,4 +229,180 @@ module.exports = function(app) {
...
+                        // the array, so this returns true if
+                        // the slug is not in slugNames
+                        if (slugNames.indexOf(obj.slugs[i]) === -1) {
+                            newSlugs.push(obj.slugs[i]);
+                        }
+                    }
+
+                    knex('projectslugs')
+                    .where({project: project.id}).then(function(existingSlugs) {
+                        existingSlugs = existingSlugs.map(function(slug) {
```
LyonesGamer: I'd suggest we comment this function a lot more, because it's difficult to read (by necessity, but still), especially keeping track of all the slugs variables (obj.slugs, slugs, existingSlugs, newSlugs, delSlugs...).

src/projects.js
```diff
@@ -229,4 +229,180 @@ module.exports = function(app) {
...
+
+                        Promise.all(slugsPromises).then(function() {
+                            if (obj.slugs.length) {
+                                project.slugs = obj.slugs;
+                            } else {
+                                project.slugs = existingSlugs;
+                            }
+
+                            project.owner = req.body.auth.user;
+                            res.send(JSON.stringify(project));
```
LyonesGamer: If we're going to have a catch block to send server errors on some database calls, we should probably be consistent and put it on all of them, even if it does bloat our codebase a little bit.

src/projects.js
```diff
@@ -229,4 +229,180 @@ module.exports = function(app) {
...
         });
     });
+
+    app.post(app.get('version') + '/projects/:slug', function(req, res) {
+        let obj = req.body.object;
+
+        // valid keys
+        let validKeys = ['name', 'uri', 'owner', 'slugs'];
+        for (let key in obj) {
+            // indexOf returns -1 if the parameter is in
```
matthewrsj: 'is not in'

tschuy: *is not in the array*

Thanks mrsj for pointing that out!

LyonesGamer:
I would definitely call this good and ready. Tests and linter pass, good coverage, does what we need it to do. No problems I can see. +1

Kennric:
I think it's ready. +1

PR #103, opened 2015-08-05
Add slug regex checks

### LINKED ISSUE #100: [3] Add slug validating regex
Valid:
            test
            a1b2-c3

          Invalid:
            1a2b
            a1b2--c3

tschuy:
This fixes #100 .

There are comments in the the regex for what it SHOULD match. If you think of a test case that should be added, please mention it.

### DIFF
--- a/docs/source/design/draft_models.rst
+++ b/docs/source/design/draft_models.rst
@@ -6,10 +6,6 @@ Draft Data Models
 
 Below are the database models to be used in TimeSync.
 
-Slugs are `as used in Django <https://docs.djangoproject.com/en/1.8/glossary/#term-slug>`_.
-That is, a valid slug consists of any combination of upper and lowercase letters, numbers,
-hyphens, and underscores.
-
 Projects:
 ---------
 
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -11,9 +11,9 @@ module.exports = {
           Valid:
             test
             a1b2-c3
+            1a2b
 
           Invalid:
-            1a2b
             a1b2--c3
 
           Based on http://stackoverflow.com/a/19256344
@@ -24,7 +24,8 @@ module.exports = {
             return false;
         }
 
-        var isSlug = new RegExp('^[a-z]+[a-z0-9]*(?:-[a-z0-9]+)*$');
+        var isSlug = new RegExp('^(?:[0-9]*-)*[0-9]*[a-z]+[a-z0-9]*' +
+            '(?:-[a-z0-9]+)*$');
         return isSlug.test(slug);
     },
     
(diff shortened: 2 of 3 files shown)


tests/test.js
```diff
@@ -59,7 +59,6 @@ describe('Endpoints', function() {
     require('./times')(expect, request, baseUrl);
     require('./activities')(expect, request, baseUrl);
     require('./projects')(expect, request, baseUrl);
-    require('./helpers')(expect);
```
Kennric: is this intentionally removed?

tschuy: Yes, it's duplicated under both Endpoints and Helpers. It's not an endpoint :P

Kennric: Ah, ok, I was thinking this was responsible for making the helpers tests run.

Kennric:
This looks good to me, regex seems reasonable thorough, and I think we have a good standard for a slug - the one question I have is, should we exclude leading integers? I am not sure if that is useful or important.

iankronquist:
What's the spec for slugs? I thought it was [the same as Django's](https://github.com/osuosl/timesync-node/blob/develop/docs/source/design/draft_models.rst#draft-data-models). If so, we should copy [Django's slug regex](https://github.com/django/django/blob/master/django/core/validators.py#L211) instead of rolling our own which matches different things.

tschuy:
@iankronquist it's https://github.com/osuosl/timesync/pull/16

ElijahCaine:
I tested the regex  (by hand). The tests pass and seem to cover all of our bases. The fact that our slug definition is a little wonky isn't ideal, but the rules are simple enough (and in the comments!). +1

ElijahCaine:
We should consider changing the regex from `^[a-z]+[a-z0-9]*(?:-[a-z0-9]+)*$``` to ``^(?:[0-9]*-)*[0-9]*[a-z]+[a-z0-9]*(?:-[a-z0-9]+)*$`` as it allows us to have preceding numbers but does not allow for the regex to just be a number.

ElijahCaine:
This new regex structure is nice and easy to read. The tests pass and they seem to cover all of our bases. Much nicer than the alternative of one giant regex monstrosity. +1

jordane:
@tschuy There are no tests for the regex that contain non-alphanumeric characters. It'd be wise to add those in case the regexes get refactored later.

ElijahCaine:
Right. I completely didn't catch that it was missing non-alphanumeric tests. These look good. Hopefully 15 tests for the regex is good enough. I can't think of any more input that we could test for. +1 (again)

ElijahCaine:
Everything *still* looks good. Post +1 from me.

Kennric:
This looks pretty thorough, I can't help feeling that our regex could be simpler, but it works, the tests test our specified slug criteria, and the regex can be streamlined later if we figure out how. +1

Issue #101, opened 2015-08-05
[5] DELETE Projects Endpoint


MaraJade:
Create tests and implementation to allow users to delete projects from the database

PR #104, opened 2015-08-05
project/:slug DELETE endpoint

MaraJade:
refs #101

### DIFF
--- a/src/projects.js
+++ b/src/projects.js
@@ -241,9 +241,7 @@ module.exports = function(app) {
                 req.params.slug);
 
         // Get times associated with project
-        knex('times').where('project', '=', function() {
-            this.table('projects').select('id').where('id', '=', projectId);
-        }).then(function(times) {
+        knex('times').where('project', '=', projectId).then(function(times) {
             // If there are times associated, return an error
             if (times.length > 0) {
                 res.set('Allow', 'GET, POST');
@@ -251,7 +249,7 @@ module.exports = function(app) {
                 return res.status(err.status).send(err);
             // Otherwise delete project
             } else {
-                knex('projects').select('id').where('id', '=', projectId)
+                knex('projects').where('id', '=', projectId)
                 .del().then(function(numObj) {
 
                     /* When deleting something from the table, the number of
--- a/src/projects.js
+++ b/src/projects.js
@@ -237,10 +237,10 @@ module.exports = function(app) {
             return res.status(err.status).send(err);
         }
 
-        // delete matching project 
+        // delete matching project
         var projectSubquery = knex('projectslugs').select('project')
         .where('name', req.params.slug);
-        knex('projects').select('id').where( 'id', '=', projectSubquery)
+        knex('projects').select('id').where('id', '=', projectSubquery)
         .del().then(function(numObj) {
 
             /* When deleting something from the table, the number of objects
--- a/src/projects.js
+++ b/src/projects.js
@@ -238,7 +238,7 @@ module.exports = function(app) {
 
         // Get project id
         let projectId = knex('projectslugs').select('project').where('name',
-                req.params.slug); 
+                req.params.slug);
 
         // Get times associated with project
         knex('times').where('project', '=', function() {
--- a/src/projects.js
+++ b/src/projects.js
@@ -230,37 +230,31 @@ module.exports = function(app) {
         });
     });
 
-    /*
     app.delete(app.get('version') + '/projects/:slug', function(req, res) {
         if (!helpers.validateSlug(req.params.slug)) {
             console.log('Invalid slug');
             var err = errors.errorInvalidIdentifier('slug', req.params.slug);
-            console.log('Invalid Identifier:', err);
             return res.status(err.status).send(err);
         }
 
         // delete matching project 
-        knex('projects').where('slug', req.params.slug).del()
-        .then(function(numObj) {
-            console.log('Inside then');
+        var projectSubquery = knex('projectslugs').select('project')
+        .where('name', req.params.slug);
+        knex('projects').select('id').where( 'id', '=', projectSubquery)
+        .del().then(function(numObj) {
+
             /* When deleting something from the table, the number of objects
                deleted is returned. So to confirm that deletion was successful,
                make sure that the number returned is at least one. */
-/*
             if (numObj >= 1) {
-                console.log('Deleted at least one object');
                 return res.send();
             }
-
-            console.log('Did not delete object');
 
             var err = errors.errorObjectNotFound('slug', req.params.slug);
             return res.status(err.status).send(err);
         }).catch(function(error) {
-            console.log('Error');
             var err = errors.errorServerError(error);
             return res.status(err.status).send(err);
         });
     });
-    */
 };
--- a/src/projects.js
+++ b/src/projects.js
@@ -232,7 +232,6 @@ module.exports = function(app) {
 
     app.delete(app.get('version') + '/projects/:slug', function(req, res) {
         if (!helpers.validateSlug(req.params.slug)) {
-            console.log('Invalid slug');
             var err = errors.errorInvalidIdentifier('slug', req.params.slug);
             return res.status(err.status).send(err);
         }
--- a/src/projects.js
+++ b/src/projects.js
@@ -229,4 +229,38 @@ module.exports = function(app) {
             return res.status(err.status).send(err);
         });
     });
+
+    /*
+    app.delete(app.get('version') + '/projects/:slug', function(req, res) {
+        if (!helpers.validateSlug(req.params.slug)) {
+            console.log('Invalid slug');
+            var err = errors.errorInvalidIdentifier('slug', req.params.slug);
+            console.log('Invalid Identifier:', err);
+            return res.status(err.status).send(err);
+        }
+
+        // delete matching project 
+        knex('projects').where('slug', req.params.slug).del()
+        .then(function(numObj) {
+            console.log('Inside then');
+            /* When deleting something from the table, the number of objects
+               deleted is returned. So to confirm that deletion was successful,
+               make sure that the number returned is at least one. */
+/*
+            if (numObj >= 1) {
+                console.log('Deleted at least one object');
+                return res.send();
+            }
+
+            console.log('Did not delete object');
+
+            var err = errors.errorObjectNotFound('slug', req.params.slug);
+            return res.status(err.status).send(err);
+        }).catch(function(error) {
+            console.log('Error');
+            var err = errors.errorServerError(error);
+            return res.status(err.status).send(err);
+        });
+    });
+    */
 };
--- a/src/projects.js
+++ b/src/projects.js
@@ -237,22 +237,10 @@ module.exports = function(app) {
         }
 
         // delete matching project
-/**
-        var projectSubquery = knex('projectslugs').select('project')
-        .where('name', req.params.slug);
-        var projectId = knex('projects').select('id').where('id', '=', projectSubquery);
-        if(knex('times').where('project', '=', projectId).length != 0) {
-            var err = errors.errorRequestFailure('project', projectSubquery);
-            return res.status(err.status).send(err);
-
-        } else {
-*/
-
-
-
         knex('times').where('project', '=', function() {
             this.table('projects').select('id').where('id', '=', function() {
-            this.table('projectslugs').select('project').where('name', req.params.slug);
+                this.table('projectslugs').select('project').where('name',
+                        req.params.slug);
             });
         }).then(function(times) {
 
@@ -261,18 +249,22 @@ module.exports = function(app) {
                 let err = errors.errorRequestFailure('project');
                 return res.status(err.status).send(err);
             } else {
-                knex('projects').select('id').where('id', '=', function() {
-                    this.table('projectslugs').select('project').where('name', req.params.slug)
+                knex('projects').select('id').where('id', '=',
+                        function() {
+                    this.table('projectslugs').select('project').where('name',
+                            req.params.slug);
                 }).del().then(function(numObj) {
 
-                    /* When deleting something from the table, the number of objects
-                       deleted is returned. So to confirm that deletion was successful,
-                       make sure that the number returned is at least one. */
+                    /* When deleting something from the table, the number of
+                       objects deleted is returned. So to confirm that deletion
+                       was successful, make sure that the number returned is at
+                       least one. */
                     if (numObj >= 1) {
                         return res.send();
                     }
 
-                    var err = errors.errorObjectNotFound('slug', req.params.slug);
+                    var err = errors.errorObjectNotFound('slug',
+                            req.params.slug);
                     return res.status(err.status).send(err);
                 }).catch(function(error) {
                     var err = errors.errorServerError(error);
--- a/src/projects.js
+++ b/src/projects.js
@@ -237,23 +237,48 @@ module.exports = function(app) {
         }
 
         // delete matching project
+/**
         var projectSubquery = knex('projectslugs').select('project')
         .where('name', req.params.slug);
-        knex('projects').select('id').where('id', '=', projectSubquery)
-        .del().then(function(numObj) {
+        var projectId = knex('projects').select('id').where('id', '=', projectSubquery);
+        if(knex('times').where('project', '=', projectId).length != 0) {
+            var err = errors.errorRequestFailure('project', projectSubquery);
+            return res.status(err.status).send(err);
 
-            /* When deleting something from the table, the number of objects
-               deleted is returned. So to confirm that deletion was successful,
-               make sure that the number returned is at least one. */
-            if (numObj >= 1) {
-                return res.send();
+        } else {
+*/
+
+
+
+        knex('times').where('project', '=', function() {
+            this.table('projects').select('id').where('id', '=', function() {
+            this.table('projectslugs').select('project').where('name', req.params.slug);
+            });
+        }).then(function(times) {
+
+            if (times.length > 0) {
+                res.set('Allow', 'GET, POST');
+                let err = errors.errorRequestFailure('project');
+                return res.status(err.status).send(err);
+            } else {
+                knex('projects').select('id').where('id', '=', function() {
+                    this.table('projectslugs').select('project').where('name', req.params.slug)
+                }).del().then(function(numObj) {
+
+                    /* When deleting something from the table, the number of objects
+                       deleted is returned. So to confirm that deletion was successful,
+                       make sure that the number returned is at least one. */
+                    if (numObj >= 1) {
+                        return res.send();
+                    }
+
+                    var err = errors.errorObjectNotFound('slug', req.params.slug);
+                    return res.status(err.status).send(err);
+                }).catch(function(error) {
+                    var err = errors.errorServerError(error);
+                    return res.status(err.status).send(err);
+                });
             }
-
-            var err = errors.errorObjectNotFound('slug', req.params.slug);
-            return res.status(err.status).send(err);
-        }).catch(function(error) {
-            var err = errors.errorServerError(error);
-            return res.status(err.status).send(err);
         });
     });
 };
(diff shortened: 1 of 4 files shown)


tests/delete.js
```diff
@@ -0,0 +1,44 @@
...
+            .then(function(project) {
+                expect(project).to.be.an('undefined');
+                //expect(res.statusCode).to.equal(404);
+                done();
+            });
+        });
+
+        it('Fails if it receives an invalid project', function(done) {
+            deleter.deleteProject('Nothing').then(function(project) {
+                //expect(res.statusCode).to.equal(400);
```
tschuy: I think it should probably throw an error with the name of the project that was being deleted.

tschuy: How is this different from the test on line 37? (This is an invalid slug, was that intentional?)

MaraJade: One is trying to delete with an invalid project, one is with an invalid slug. Are those too similar to need separate tests?

tschuy: I think you should use a more obviously invalid slug than 'Nothing'.

MaraJade: ...I couldn't come up with an invalid project

tests/delete.js
```diff
@@ -0,0 +1,44 @@
...
+            deleter.deleteProject('Whats Fresh')
+            .then(function(project) {
+                expect(project).to.be.an('undefined');
+                //expect(res.statusCode).to.equal(404);
+                done();
+            });
+        });
+
+        it('Fails if it receives an invalid project', function(done) {
+            deleter.deleteProject('Fail Project').then(function(project) {
```
tschuy: I would use something with copious amounts of punctuation like ``$(@#project`` or something ike that.

tests/delete.js
```diff
@@ -0,0 +1,44 @@
+// delete endpoints
+var deleter = require('../src/delete');
+
+module.exports = function(expect) {
+    /* DELETE one of the projects endpoints and check whether it can still
+       be retrieved from the database */
+    describe('DELETE /projects', function() {
+        it('Deletes the desired project', function(done) {
+            deleter.deleteProject('Whats Fresh')
```
tschuy: this is an endpoint, not a helper function, so this should look more like the tests in, say, ``tests/activities.js``.

Sorry about not catching that earlier.

tschuy:
I'm confused what your ``DELETE /projects`` does that's different from ``DELETE /projects/:slug``.

Also, these should be endpoint tests, so usnig the ``request`` library to DELETE the object.

MaraJade:
I removed the apparently redundant test and attempted to use the request library. Hopefully it's a better test now.

src/app.js
```diff
@@ -37,6 +37,8 @@ require('./projects')(app);
 require('./activities')(app);
 require('./times')(app);
 
+module.exports = app
```
tschuy: Needs a semicolon.

src/delete.js
```diff
@@ -0,0 +1,9 @@
+// delete endpoints
+
+var app = require('./app');
+var knex = app.get('knex');
+
+module.exports = {
```
tschuy: Model this after one of the endpoint files, with it as a function that takes app as an argument.

tests/delete.js
```diff
@@ -0,0 +1,51 @@
+// del endpoints
+var deleter = require('../src/delete');
```
tschuy: Since this is an endpoint test you don't need to require this.

tests/delete.js
```diff
@@ -0,0 +1,51 @@
...
+
+module.exports = function(expect, request) {
+    /* del one of the projects endpoints and check whether it can still
+       be retrieved from the database */
+    describe('del /projects', function() {
```
tschuy: ``describe('DELETE /projects'`` and ``it('deletes the desired project'``.

tests/delete.js
```diff
@@ -0,0 +1,51 @@
...
+        it('dels the desired project', function(done) {
+            request.del(baseUrl + 'projects/whatsfresh', function(err, res, body){
```
tschuy: No project with this slug exists.

tests/delete.js
```diff
@@ -0,0 +1,49 @@
...
+    /* delete one of the projects endpoints and check whether it can still
+       be retrieved from the database */
+    describe('DELTE /projects', function() {
+        it('deletes the desired project', function(done) {
+            request.del(baseUrl + 'projects/wf', function(err, res, body){
+                var jsonBody = JSON.parse(body);
+                var expectedResult = {}
+                expect(res.statusCode).to.equal(200);    
+                request.get(baseUrl + 'projects/wf', function(err, res, body){
+                    var jsonBody = JSON.parse(body);
```
tschuy: We shouldn't have a body at all. I would get rid of expectedResult, jsonBody, and on line 19, ``expect(res.body).to.be.a.('null')``.

tests/delete.js
```diff
@@ -0,0 +1,49 @@
...
+                        error: 'Object not found',
+                        text: 'Nonexistent project'
+                    }
+
+                    expect(jsonBody).to.deep.equal(expectedResult);
+                    expect(res.statusCode).to.equal(404);
+                    done();
+                });
+            });
+            /**
```
tschuy: Get rid of this block comment

src/delete.js
```diff
@@ -0,0 +1,7 @@
+// delete endpoints
```
tschuy: Get rid of this file

tests/test.js
```diff
@@ -59,7 +59,7 @@ describe('Endpoints', function() {
     require('./times')(expect, request, baseUrl);
     require('./activities')(expect, request, baseUrl);
     require('./projects')(expect, request, baseUrl);
-    require('./helpers')(expect);
+    require('./delete')(expect);
```
tschuy: This file no longer exists, so this require is unnecessary.

tschuy:
After you rebase this, it should be good to +1

src/app.js
```diff
@@ -37,6 +37,8 @@ require('./projects')(app);
 require('./activities')(app);
 require('./times')(app);
 
+module.exports = app;
```
tschuy: What re you using app for here?

tests/projects.js
```diff
@@ -99,4 +99,33 @@ module.exports = function(expect, request, baseUrl) {
...
+
+                    expect(body).to.be.a(null);
+                    expect(res.statusCode).to.equal(404);
+                    done();
+                });
+            });
+        });
+
+        it('Fails if it receives an invalid project', function(done) {
+            request.del('Not.a-project!').then(function(project) {
```
LyonesGamer: Shouldn't this be `request.del(baseUrl + 'projects/Not.a.project!')`? That doesn't even have a proper hostname.

tschuy: it should indeed. my reviewing bad!

tests/projects.js
```diff
@@ -99,4 +99,33 @@ module.exports = function(expect, request, baseUrl) {
...
+                var jsonBody = JSON.parse(body);
+                var expectedResult = {
+                    status: 404,
```
LyonesGamer: Also, DELETE /projects/Not.a.project! should return an Invalid Identifier error because that's not a valid slug. Object Not Found is for something that *could* be a slug but isn't.

tests/projects.js
```diff
@@ -99,4 +99,33 @@ module.exports = function(expect, request, baseUrl) {
...
+                });
+            });
+        });
+
+        it('Fails if it receives an invalid project', function(done) {
+            request.del(baseUrl + 'projects/Not.a-project!', function(err, res, body) {
+                var jsonBody = JSON.parse(body);
+                var expectedResult = {
+                    status: 404,
+                    error: 'Object not found',
```
tschuy: This should be an invalid identifier error, and there should be another test that tests a valid slug that doesn't exist.

tests/projects.js
```diff
@@ -99,4 +99,48 @@ module.exports = function(expect, request, baseUrl) {
...
+                var expectedResult = {
+                    status: 400,
+                    error: 'The provided identifier was invalid',
+                    text: 'Expected notaproject but recieved Not.a-project!',
+                    values: ['Not.a-project!']
+                };
+
+                expect(res.statusCode).to.equal(400);
+                expect(body).to.deep.equal(expectedResult);
+                done();
```
tschuy: can you test that this actually doesn't delete anything by checking the ``GET /projects`` endpoint?

tests/projects.js
```diff
@@ -396,4 +397,77 @@ module.exports = function(expect, request, baseUrl) {
...
+                            slugs: ['wf'],
+                            owner: 'tschuy',
+                            id: 3
+                        }
+                    ];
+
+                    expect(res1.statusCode).to.equal(200);
+                    expect(jsonBody).to.deep.equal(expectedResult1);
+                    expect(res.statusCode).to.equal(400);
+                    expect(body).to.deep.equal(expectedResult);
```
tschuy: I would just reuse the variable names err, res, and body. They'll take precedence over  the other ones and make the variables you're using in general less confusing.

You can ``expect`` the values for the del request outside of this inner get request.

tests/projects.js
```diff
@@ -396,4 +397,77 @@ module.exports = function(expect, request, baseUrl) {
...
+        it('Fails if it receives an invalid project', function(done) {
+            request.del(baseUrl + 'projects/Not.a-project!', function(err, res, body) {
+                var expectedResult = {
+                    status: 400,
+                    error: 'The provided identifier was invalid',
+                    text: 'Expected slug but received Not.a-project!',
+                    values: [ 'Not.a-project!' ]
+                };
+
+                //console.log('Test invalid project:', body, expectedResult, err);
```
tschuy: debugging comment

tests/helpers.js
```diff
@@ -89,6 +89,11 @@ module.exports = function(expect, app) {
             done();
         });
 
+        it('returns false with non-alphanumeric input', function(done) {
```
tschuy: looks like you duplicated this section

tests/helpers.js
```diff
@@ -121,7 +126,14 @@ module.exports = function(expect, app) {
 
         it('throws when passed a null slug', function(done) {
             helpers.checkProject(null).then().catch(function(err) {
-                expect(err).to.deep.equal({type: 'invalid', value: null});
```
tschuy: This line should still be here

tests/projects.js
```diff
@@ -396,4 +397,78 @@ module.exports = function(expect, request, baseUrl) {
         });
     });
 
+    describe('DELETE /projects/:slug', function() {
+        it('deletes the desired project', function(done) {
+            request.del(baseUrl + 'projects/wf', function(err, res){
+                expect(res.statusCode).to.equal(200);    
+                request.get(baseUrl + 'projects/wf', function(err, res, body){
+
+                    expect(body).to.be.a(null);
```
tschuy: We should expect an object not found error http://timesync.readthedocs.org/en/latest/draft_errors.html#object-not-found

tests/projects.js
```diff
@@ -396,4 +397,78 @@ module.exports = function(expect, request, baseUrl) {
...
+                var jsonBody = JSON.parse(body)
+                var expectedResult = {
+                    status: 404,
+                    error: 'Object not found',
+                    text: 'Nonexistant project'
+                };
+
+                expect(res.statusCode).to.equal(404);
+                expect(jsonBody).to.deep.equal(expectedResult);
+                done();
```
tschuy: can you test that ``GET projects/`` still has teh same stuff like you did above?

tests/helpers.js
```diff
@@ -133,5 +133,13 @@ module.exports = function(expect, app) {
                 done();
             });
         });
+
```
tschuy: This test is still duplicated

tschuy:
These get a +1 from me. @osuosl/devs another review?

tests/projects.js
```diff
@@ -396,4 +397,114 @@ module.exports = function(expect, request, baseUrl) {
...
+                            uri: 'https://github.com/osu-cass/whats-fresh-api',
+                            name: 'Whats Fresh',
+                            slugs: ['wf'],
+                            owner: 'tschuy',
+                            id: 3
+                        }
+                    ];
+
+                    expect(res.statusCode).to.equal(200);
+                    expect(jsonBody).to.deep.equal(expectedResult);
```
athai: Should be: expect(jsonBody).to.deep.have.same.members(expectedResult) when comparing lists

tests/projects.js
```diff
@@ -396,4 +397,114 @@ module.exports = function(expect, request, baseUrl) {
...
```
athai: expect(jsonBody).to.deep.have.same.members(expectedResult)

tests/projects.js
```diff
@@ -396,4 +397,114 @@ module.exports = function(expect, request, baseUrl) {
...
+
+                request.get(baseUrl + 'projects/wf', function(err, res, body){
+
+                    var expectedResult = {
+                        status: 404,
+                        error: 'Object not found',
+                        text: 'Nonexistent project'
+                    }
+
+                    expect(body).to.deep.equal(expectedResult);
```
athai: Shouldn't you JSON.parse(body) before comparing to expectedResult?

tests/projects.js
```diff
@@ -396,4 +397,114 @@ module.exports = function(expect, request, baseUrl) {
...
+            });
+        });
+
+        it('Fails if it receives an non-existant project', function(done) {
+            request.del(baseUrl + 'projects/doesntexist', function(err, res, body) {
+                var jsonBody = JSON.parse(body)
+                var expectedResult = {
+                    status: 404,
+                    error: 'Object not found',
+                    text: 'Nonexistant project'
```
athai: You spelled NonexistEnt wrong.  The tests won't pass unless it's spelled correctly.

MaraJade:
Everything should be implemented and working now. Linter isn't passing because a url is too long for its indentation

tschuy:
And a few other reasons :P

You can split strings across new lines by using ' + ':

```
function() {
    console.log('I really like long strings, and I think that every string ' + 
        'should be a really long string'); 
}
```

MaraJade:
I actually fixed them this time.

src/projects.js
```diff
@@ -229,4 +229,32 @@ module.exports = function(app) {
             return res.status(err.status).send(err);
         });
     });
+
+    app.delete(app.get('version') + '/projects/:slug', function(req, res) {
+        if (!helpers.validateSlug(req.params.slug)) {
+            console.log('Invalid slug');
```
tschuy: left over comment

tschuy:
This will need a new test that tests that things don't get deleted if they're still referenced, and the existing tests should be changed so that they use a project that's not referenced.

tests/projects.js
```diff
@@ -396,4 +396,136 @@ module.exports = function(expect, request, baseUrl) {
         });
     });
 
+    describe('DELETE /projects/:slug', function() {
+        it('deletes the desired project if no times are associated with it',
+                function(done) {
+            request.del(baseUrl + 'projects/wf', function(err, res) {
```
tschuy: this will need to be changed to a project that doesn't have times associated with it.

tests/projects.js
```diff
@@ -396,4 +396,136 @@ module.exports = function(expect, request, baseUrl) {
...
+                    expect(jsonBody).to.deep.equal(expectedResult);
+                    expect(res.statusCode).to.equal(404);
+                    done();
+                });
+            });
+        });
+
+        it('Failes if it recieves a project with times assiciated',
+            function(done) {
+                request.del(baseUrl + 'projects/pgd',
```
tschuy: pgd has no times associated.

tests/projects.js
```diff
@@ -396,4 +396,136 @@ module.exports = function(expect, request, baseUrl) {
...
+                    };
+
+                    expect(jsonBody).to.deep.equal(expectedResult);
+                    expect(res.statusCode).to.equal(404);
+                    done();
+                });
+            });
+        });
+
+        it('Failes if it recieves a project with times assiciated',
```
tschuy: associated is misspelled.

tests/projects.js
```diff
@@ -396,4 +396,136 @@ module.exports = function(expect, request, baseUrl) {
...
+            function(done) {
+                request.del(baseUrl + 'projects/pgd',
+                    function(err, res, body) {
+                        var expectedResult = {
+                            status: 405,
+                            error: 'DELETE not allowed',
+                            text: 'Project has associated time'
```
tschuy: you should make the error and text match https://github.com/osuosl/timesync/pull/20/files#diff-459f373b6fe1c0684fe24c353ae5ee62R227

tests/projects.js
```diff
@@ -396,4 +396,136 @@ module.exports = function(expect, request, baseUrl) {
...
+                    expect(jsonBody).to.deep.have.same.members(expectedResult);
+                });
+
+                expect(res.statusCode).to.equal(400);
+                expect(jsonBody).to.deep.equal(expectedResult);
+                done();
+            });
+        });
+
+        it('Fails if it receives an non-existant project',
```
tschuy: Find and replace "existant" -> "existent"

tests/projects.js
```diff
@@ -396,4 +396,138 @@ module.exports = function(expect, request, baseUrl) {
...
+                    };
+
+                    expect(jsonBody).to.deep.equal(expectedResult);
+                    expect(res.statusCode).to.equal(404);
+                    done();
+                });
+            });
+        });
+
+        it('Failes if it recieves a project with times associated',
```
matthewrsj: 'Fails'

tschuy:
lgtm, now fix the implementation :)

src/projects.js
```diff
@@ -229,4 +229,48 @@ module.exports = function(app) {
...
         });
     });
+
+    app.delete(app.get('version') + '/projects/:slug', function(req, res) {
+        if (!helpers.validateSlug(req.params.slug)) {
+            var err = errors.errorInvalidIdentifier('slug', req.params.slug);
+            return res.status(err.status).send(err);
+        }
+
+        // delete matching project
```
tschuy: I'd update this comment to reflect that this actually gets any times with the project.

src/projects.js
```diff
@@ -229,4 +229,48 @@ module.exports = function(app) {
...
+        knex('times').where('project', '=', function() {
+            this.table('projects').select('id').where('id', '=', function() {
```
tschuy: since you're using this query twice, you might want to consider moving it into its own variable like https://github.com/osuosl/timesync-node/blob/develop/src/projects.js#L309

tschuy:
+1

src/projects.js
```diff
@@ -229,4 +229,47 @@ module.exports = function(app) {
...
+            return res.status(err.status).send(err);
+        }
+
+        // Get project id
+        let projectId = knex('projectslugs').select('project').where('name',
+                req.params.slug);
+
+        // Get times associated with project
+        knex('times').where('project', '=', function() {
+            this.table('projects').select('id').where('id', '=', projectId);
```
LyonesGamer: I'm not sure I see the necessity of this line. You're requesting the project ID of of a project...with a given project ID. The result of the call is going to be identical to the one returned from `projectId`. Just do `knex('times').where('project', '=', projectId);`

src/projects.js
```diff
@@ -229,4 +229,47 @@ module.exports = function(app) {
...
+        }).then(function(times) {
+            // If there are times associated, return an error
+            if (times.length > 0) {
+                res.set('Allow', 'GET, POST');
+                let err = errors.errorRequestFailure('project');
+                return res.status(err.status).send(err);
+            // Otherwise delete project
+            } else {
+                knex('projects').select('id').where('id', '=', projectId)
```
LyonesGamer: You don't need the select call here. I'm not even entirely sure what the result would be of calling select() and del() at the same time...

LyonesGamer:
+1, even though it claims a test failed. Looks like just a weird bug.

tschuy:
+1

PR #105, opened 2015-08-06
checking things work the way I think

iankronquist:


iankronquist:
Editing from the GitHub web interface works the way I think it does.

PR #107, opened 2015-08-06
Thai/delete times

### LINKED ISSUE #94: [5] DELETE Times Endpoint
Create tests and implementation to allow the user the user to delete times from the database.

athai:
Tests for deleting time endpoints. Fixes #94.


tests/delete.js
```diff
@@ -0,0 +1,36 @@
+// delete endpoints
+var deleter = require('../src/delete');
```
tschuy: we don't require endpoint files. The test server is set up in ``test.js``, and then we test against that.

tests/delete.js
```diff
@@ -0,0 +1,36 @@
...
+
+module.exports = function(expect, request, baseUrl) {
+    /* DELETE one of the times endpoints and check whether it can still be
+       retrieved from the database */
+    describe('DELETE /times/:id', function() {
+        it('Deletes the desired time instance', function(done) {
+            deleter.deleteTime(1).then(function(time) {
```
tschuy: ``request.del``, not ``deleter``. These are endpoint tests, so we need to request the endpoint and check what we get back.

tests/delete.js
```diff
@@ -0,0 +1,49 @@
...
+                expect(res.statusCode).to.equal(404);
+                done();
+            });
+        });
+
+        // Checks that an invalid time id will fail /ex: time id = 'tabby'
+        it('Fails if it receives an invalid time id', function(done) {
+            request.del(baseUrl + 'times/tabby', function(err, res) {
+                // ('notanid').then(function(timeid) {
+                expect(timeid).to.be.an('undefined');
```
tschuy: timeid isn't a variable here.

tests/delete.js
```diff
@@ -0,0 +1,49 @@
...
+                expect(res.statusCode).to.equal(400);
```
tschuy: also check the body of the request. 

both of these comments apply to the other fail check.

tests/test.js
```diff
@@ -59,6 +59,8 @@ describe('Endpoints', function() {
     require('./times')(expect, request, baseUrl);
     require('./activities')(expect, request, baseUrl);
     require('./projects')(expect, request, baseUrl);
+    require('./delete')(expect);
```
tschuy: this file doesn't need to be required anymore, since it's deleted. (hah!)

tests/delete.js
```diff
@@ -0,0 +1,49 @@
+//var deleter = require('../src/delete');
```
tschuy: These tests should go in the ``tests/times.js`` file

tests/test.js
athai: Isn't that line so that new tests will be included when you run npm test?

tschuy: Right, but the tests shouldn't be in ``delete.js``. They should be in ``times.js``.

tests/times.js
```diff
@@ -90,4 +90,63 @@ module.exports = function(expect, request, baseUrl) {
...
+
+    /* DELETE one of the times endpoints and check whether it can still be
+       retrieved from the database */
+
+    describe('DELETE /times/:id', function() {
+        it('deletes the desired time instance', function(done) {
+            request.del(baseUrl + 'times/1', function(err, res) {
+                expect(err).to.be.a('null');
+                expect(res.statusCode).to.equal(200);
+                expect(time).to.be.an('undefined');
```
Kennric: I'm unclear where 'time' is coming from here, or why we are making sure it hasn't come from anywhere (undefined). We didn't ask for this variable in the request, nor define it.

tests/times.js
```diff
@@ -90,4 +90,63 @@ module.exports = function(expect, request, baseUrl) {
...
+                    };
+
+                    expect(jsonBody).to.have.same.members(expectedError);
+                    expect(res.statusCode).to.equal(404);
+                    done();
+                });
+            });
+        });
+
+        // Checks that a nonexistent time id will fail /ex: time id = 6013
```
Kennric: It may be a good idea to make sure that -nothing- gets deleted here, just in case some bug might try deleting other things when given a bad id. So another times request and comparison to expected data might be a good idea in this and the next test.

tests/times.js
```diff
@@ -90,4 +90,63 @@ module.exports = function(expect, request, baseUrl) {
...
+        it('fails if it receives a bad time id', function(done) {
+            request.del(baseUrl + 'times/6013', function(err, res) {
+                var expectedError = {
+                    status: 404,
+                    error: 'Object not found',
+                    text: 'Nonexistent time id'
+                };
+
+                expect(6013).to.be.an('undefined');
```
Kennric: isn't 6013 just an int in this context? An int isn't undefined.

tests/test.js
```diff
@@ -59,6 +59,7 @@ describe('Endpoints', function() {
     require('./times')(expect, request, baseUrl);
     require('./activities')(expect, request, baseUrl);
     require('./projects')(expect, request, baseUrl);
+
```
tschuy: random newline?

tests/times.js
```diff
@@ -90,4 +90,114 @@ module.exports = function(expect, request, baseUrl) {
...
+                            created_at: null,
+                            updated_at: null,
+                            activities : ['docs', 'dev']
+                            // jscs: enable
+                        }
+                    ];
+
+                    expect(err).to.be.a('null');
+                    expect(res.statusCode).to.equal(200);
+                    expect(jsBody).to.deep.equal(expectedResult);
```
tschuy: You'll need this to ``.deep.have.same.members()`` to compare lists.

tests/times.js
```diff
@@ -90,4 +90,114 @@ module.exports = function(expect, request, baseUrl) {
...
+                            created_at: null,
+                            updated_at: null,
+                            activities: ['docs', 'dev']
+                            // jscs: enable
+                        }
+                    ];
+
+                    expect(err).to.be.a('null');
+                    expect(res.statusCode).to.equal(200);
+                    expect(jsBody).to.deep.equal(expectedResult);
```
tschuy: same here.

tschuy:
Looks good to me, get one more +1 and then you can implement :)

ElijahCaine:
These tests seem solid. +1 from me

tests/times.js
```diff
@@ -90,4 +90,114 @@ module.exports = function(expect, request, baseUrl) {
...
+                    expect(res.statusCode).to.equal(404);
+                    done();
+                });
+            });
+        });
+
+        // Checks that a nonexistent time id will fail /ex: time id = 6013
+        it('fails if it receives a nonexistent time id', function(done) {
+            request.del(baseUrl + 'times/6013', function(err, res, body) {
+                var jsonBody = JSON.parse(body);
```
ElijahCaine: ``body`` is already a JSON object so you are unable to ``JSON.parse(body)``
iirc we actually do this a few other times in other tests. The test runner's output looks like this:
```
  2) Endpoints DELETE /times/:id fails if it receives a nonexistent time id:
     Uncaught SyntaxError: Unexpected token C
      at Object.parse (native)
      at Request._callback (tests/times.js:122:37)
      at _stream_readable.js:908:16
```

tests/times.js
```diff
@@ -90,4 +90,114 @@ module.exports = function(expect, request, baseUrl) {
...
+
+                    done();
+                });
+            });
+        });
+
+        // Checks that an invalid time id will fail /ex: time id = 'tabby'
+        it('fails if it receives an invalid time id', function(done) {
+            request.del(baseUrl + 'times/tabby', function(err, res, body) {
+                var jsonBody = JSON.parse(body);
```
ElijahCaine: ``JSON.parse(body)`` -> ``body`` since body is already a JSON object.

tschuy: I believe for errors, express doesn't automatically set the content-type to JSON, so request doesn't automatically convert it to JSON. Notice that we need to convert it to json in ``delete-activities`` for the tests to pass.

tests/times.js
```diff
@@ -90,4 +90,114 @@ module.exports = function(expect, request, baseUrl) {
...
+                    status: 400,
+                    error: 'The provided identifier was invalid',
+                    text: 'Expected id but received tabby'
+                };
+
+                expect(jsonBody).to.deep.equal(expectedError);
+                expect(res.statusCode).to.equal(400);
+
+                request.get(baseUrl + 'times', function(err, res, body) {
+                    var jsBody = JSON.parse(body);
```
ElijahCaine: ``body`` is already a JSON object.

tests/times.js
```diff
@@ -90,4 +90,114 @@ module.exports = function(expect, request, baseUrl) {
...
+                    status: 404,
+                    error: 'Object not found',
+                    text: 'Nonexistent time id'
+                };
+
+                expect(jsonBody).to.deep.equal(expectedError);
+                expect(res.statusCode).to.equal(404);
+
+                request.get(baseUrl + 'times', function(err, res, body) {
+                    var jsBody = JSON.parse(body);
```
ElijahCaine: ``body`` is already a JSON object.

tests/times.js
```diff
@@ -90,4 +90,114 @@ module.exports = function(expect, request, baseUrl) {
...
+
+    describe('DELETE /times/:id', function() {
+        it('deletes the desired time instance', function(done) {
+            request.del(baseUrl + 'times/1', function(err, res) {
+                expect(err).to.be.a('null');
+                expect(res.statusCode).to.equal(200);
+
+                // Check if time instance was deleted from db
+                request.get(baseUrl + 'times/1', function(err, res, body) {
+                    var jsonBody = JSON.parse(body);
```
ElijahCaine: ``body`` is already a JSON object.

LyonesGamer:
Since we agreed to leave dealing with soft deletes for next sprint, I don't see anything wrong with the code. Tests and linter pass, good coverage, and it seems to work like it should. +1

tschuy:
We've decided you can't delete time entries.

PR #108, opened 2015-08-07
Thai/delete activities

### LINKED ISSUE #93: [5] DELETE Activities Endpoint
Create tests and implementation to allow the user the user to delete activities from the database.

athai:
Tests for deleting activity endpoints. Fixes #93.


tests/delete.js
```diff
@@ -0,0 +1,48 @@
...
+                    done();
+                });
+            });
+        });
+
+        // Checks that delete will fail w/ nonexistent slug
+        it('Fails if it receives a bad slug', function(done) {
+            request.del(baseUrl + 'activities/naps', function(err, res) {
+                expect('naps').to.be.an('undefined');
+                expect(res.statusCode).to.equal(404);
```
tschuy: check response body of any failure requests.

tests/delete.js
```diff
@@ -0,0 +1,48 @@
+module.exports = function(expect, request, baseUrl) {
```
tschuy: these should be in ``activities.js``.

tests/activities.js
```diff
@@ -82,4 +82,70 @@ module.exports = function(expect, request, baseUrl) {
...
+    /* DELETE one of the activities endpoints and check whether it can still
+       be retrieved from the database */
+
+    describe('DELETE /activities/:slug', function() {
+        it('Deletes the activity affiliated with the activity slug',
+        function(done) {
+            request.del(baseUrl + 'activity/docs', function(err, res) {
+                expect(err).to.be.a('null');
+                expect(res.statusCode).to.equal(200);
+                //expect('docs').to.be.an('undefined');
```
tschuy: get rid of this comment

tschuy:
:+1:

tests/activities.js
```diff
@@ -82,4 +82,69 @@ module.exports = function(expect, request, baseUrl) {
...
+                    expect('docs').to.be.an('undefined');
+                    expect(jsonBody).to.deep.have.same.members(expectedError);
+                    expect(res.statusCode).to.equal(404);
+                    done();
+                });
+            });
+        });
+
+        // Checks that delete will fail w/ nonexistent slug
+        it('fails if it receives a bad slug', function(done) {
```
ElijahCaine: bad -> nonexistent. Bad slug implies that the slug is malformation ( ``1--a`` or something like that ).

tests/test.js
```diff
@@ -59,6 +59,7 @@ describe('Endpoints', function() {
     require('./times')(expect, request, baseUrl);
     require('./activities')(expect, request, baseUrl);
     require('./projects')(expect, request, baseUrl);
+
```
ElijahCaine: Random newline?

tests/activities.js
```diff
@@ -82,4 +82,69 @@ module.exports = function(expect, request, baseUrl) {
...
+    describe('DELETE /activities/:slug', function() {
+        it('deletes the activity affiliated with the activity slug',
+        function(done) {
+            request.del(baseUrl + 'activity/docs', function(err, res) {
+                expect(err).to.be.a('null');
+                expect(res.statusCode).to.equal(200);
+
+                // Checks to see that the activity has been deleted from the db
+                request.get(baseUrl + 'activities/docs',
+                function(err, res, body) {
```
ElijahCaine: I checked our style guide and it does not specify if we indent lines like this. I think we should. @Kennric @tschuy @LyonesGamer: thoughts?

ElijahCaine:
The tests look pretty solid. I commented on a few lines about formatting and word choice, but other than that it looks good to me.

tests/activities.js
tschuy: oh, yeah, this should be indented to the same place as baseUrl, I think. That'd be pretty ugly, I wonder if there's a better way around this?

src/activities.js
```diff
@@ -40,4 +42,28 @@ module.exports = function(app) {
         });
 
     });
+
+    // DELETE implementation starts here -- delete this comment later
```
tschuy: "delete this comment later"? :stuck_out_tongue_closed_eyes:

src/activities.js
```diff
@@ -40,4 +42,28 @@ module.exports = function(app) {
...
+    app.delete(app.get('version') + '/activities/:slug', function(req, res) {
+        if (!helpers.validateSlug(req.params.slug)) {
+            var err = errors.errorInvalidIdentifier('slug', req.params.slug);
+            return res.status(err.status).send(err);
+        }
+
+        // delete matching activity
+        knex('activities').where('slug', req.params.slug).del()
+        .then(function(numObj) {
+            // if the number of objects deleted is at least one
```
tschuy: Maybe add a comment as to why this works.

tests/activities.js
```diff
@@ -83,4 +83,67 @@ module.exports = function(expect, request, baseUrl) {
...
+                    var jsonBody = JSON.parse(body);
+                    var expectedError = {
+                        status: 404,
+                        error: 'Object not found',
+                        text: 'Nonexistent activity'
+                    };
+
+                    //expect('docs').to.be.an('undefined');
```
tschuy: there are some extra comments dotted around this file

tests/activities.js
```diff
@@ -83,4 +83,119 @@ module.exports = function(expect, request, baseUrl) {
...
+
+                    done();
+                });
+            });
+        });
+
+        // Checks that delete will fail w/ nonexistent slug
+        it('fails if it receives a nonexistent slug', function(done) {
+            request.del(baseUrl + 'activities/naps', function(err, res, body) {
+                var jsonBody = JSON.parse(body);
```
ElijahCaine: This line is actually incorrect. It returns an error in the tests:
```
  2) Endpoints DELETE /times/:id fails if it receives a nonexistent time id:
     Uncaught SyntaxError: Unexpected token C
      at Object.parse (native)
      at Request._callback (tests/times.js:122:37)
      at _stream_readable.js:908:16
```
You don't ``JSON.parse(body)`` because body is already a JSON object.

src/activities.js
```diff
@@ -40,4 +42,44 @@ module.exports = function(app) {
...
+        let activityId = knex('activities').select('id')
+        .where('slug', req.params.slug);
+
+        // Check timesactivities to see if an activity (id) is being referenced
+        knex('timesactivities').where('id', '=', activityId)
+        .then(function(activity) {
+            /* If the length of the activity array is greater than 0, then
+               the activity id is being referenced by timesactivities */
+            if (activity.length > 0) {
+                res.setHeader('Allow: ', 'GET, POST');
```
LyonesGamer: Should be `set`, not `setHeader`, and the first string is just the key name(`'Allow'`, not `'Allow: '`).

src/activities.js
```diff
@@ -40,4 +42,44 @@ module.exports = function(app) {
...
+        if (!helpers.validateSlug(req.params.slug)) {
+            let err = errors.errorInvalidIdentifier('slug', req.params.slug);
+            return res.status(err.status).send(err);
+        }
+
+        let activityId = knex('activities').select('id')
+        .where('slug', req.params.slug);
+
+        // Check timesactivities to see if an activity (id) is being referenced
+        knex('timesactivities').where('id', '=', activityId)
```
LyonesGamer: I could be wrong but I'm pretty sure you need to call `select()`. So `knex('timesactivities').select('id').where(...)`

Kennric:
This looks good, my only quibble is that we should test to ensure that the Allow header is being returned. I think we approved the tests before we documented the Allow.

LyonesGamer:
Aside from the few small comments I made, and Ken's point about the Allow header, it otherwise looks good. Conditional +1 once those things are fixed.

LyonesGamer:
Looks good to me. +1

tschuy:
+1

Issue #102, opened 2015-08-05
[5] POST Times Endpoint


MaraJade:
Create tests and implementation that allows the user to update/edit an existing times object. Users should be able to POST to /times/ with a JSON body that only contains the part that is being updated.

PR #110, opened 2015-08-10
POST times/:id

ElijahCaine:
- Skeleton code for each test exist.
- I am trying to write a single testing function that I can call so I
don't duplicate code over ~30 tests (yes, it's going to be almost 30
tests).  Hopefully that pans out.
- linter does not pass
- refs #102


tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+            duration:    12,
+            user:        'tschuy',
+            project:     ['wf'],
+            notes:       '',
+            activities:  ['docs', 'dev'],
+            // jscs:disable
+            issue_uri:
+               'https:://github.com/osu-cass/whats-fresh-api/issues/56',
+            date_worked: null,
+            created_at:  null,
```
Kennric: should this be null?

tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+            notes:       'Now this is a note',
+            // jscs:disable
+            issue_uri:   'https://github.com/osuosl/pgd/pull/19',
+            date_worked: '2015-04-28',
+            // jscs:enable
+        };
+
+        // Individual pieces of the above JSON object
+        // Used for sending individual patches
+        patchedTimeDuration   = {duration:    patchedTime.duration};
```
tschuy: Why not just set the object to ``{duration:    patchedTime.duration}`` directly?

tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+        // jscs:disable
+        invalidTimeIssueUri   = {issue_uri:   invalidTime.issue_uri};
+        invalidTimeDateWorked = {date_worked: invalidTime.date_worked};
+        // jscs:enable
+        invalidTimeKey        = {key:         invalidTime.key};
+
+        var postArg = {
+            auth: {
+                user: 'tschuy',
+                password: '$2a$10$6jHQo4XTceYyQ/SzgtdhleQqkuy2G27omuIR8M' +
```
tschuy: This should be ``password``, and we should probably set up tests for authentication requirements on all of these endpoints.

tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+         * times over.
+         */
+        function sendDataToEndpoint(done,
+                                    postObj,
+                                    expectedResults,
+                                    error,
+                                    statusCode,
+                                    postBodies) {
+            postArg.object = copyJsonObject(postObj);
+            requestOptions.form = copyJsonObject(postArg);
```
tschuy: You don't need to copy here.

tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+         * Without this function you would see this exact code pretty 26
+         * times over.
+         */
+        function sendDataToEndpoint(done,
+                                    postObj,
+                                    expectedResults,
+                                    error,
+                                    statusCode,
+                                    postBodies) {
+            postArg.object = copyJsonObject(postObj);
```
tschuy: You don't need to copy here -- we're not worried about mutating postObj, we're worried about mutating postArg and requestOptions, both of which you **are** doing here.

tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+
+            // make a given post request
+            // check the error
+            // check the statusCode
+            // Also check the body of the request
+            request.post(requestOptions, function(err, res, body) {
+                expect(err).to.be.a(error);
+                expect(res.statusCode).to.equal(statusCode);
+
+                if (postBody !== undefined) {
```
tschuy: where did postBody come from?

tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+         * the tests.
+         * Without this function you would see this exact code pretty 26
+         * times over.
+         */
+        function sendDataToEndpoint(done,
+                                    postObj,
+                                    expectedResults,
+                                    error,
+                                    statusCode,
+                                    postBodies) {
```
tschuy: why is postBodies plural?

tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+        /*
+         * Okay so here's the deal.
+         * This endpoint has ~26 tests, which are honestly just 3 tests
+         * repeated 7 or 8 times (with a few exceptions).
+         * This function in theory gets rid of a lot of the repeated code in
+         * the tests.
+         * Without this function you would see this exact code pretty 26
+         * times over.
+         */
+        function sendDataToEndpoint(done,
```
tschuy: Perhaps a more explanatory name like ``checkPostToEndpoint()`` that makes it clear it actually checks stuff? Would make reading things that use it easier.

tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+            statusCode = 200;
+            postBody = undefined;
+
+            sendDataToEndpoint(done, postObj, expectedResults, error,
+                               statusCode, postBody);
+        });
+
+        // Tests valid duration field
+        it('successfully patches time with valid duration', function(done) {
+            postObj = copyJsonObject(patchedTimeDuration);
```
tschuy: Here's where I'm talking about -- why define ``patchedTimeDuration`` above if this is the only place you use it?

tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+            expectedResults = copyJsonObject(originalTime);
+            expectedResults.duration = patchedTime.duration;
+            error = 'null';
+            statusCode = 200;
+            postBody = undefined;
```
tschuy: so what exactly is ``postBody``?

tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+                }
+
+                // Always checks for valid get request
+                // err is always 'null'
+                // res.statusCode is always 200
+                // body always equals expectedresults
+                request.get(requestOptions.url, function(err, res, body) {
+                    expect(err).to.be.a('null');
+                    expect(res.statusCode).to.equal(200);
+                    console.log(body);
```
tschuy: ``console.log``

tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+            postObj = copyJsonObject(invalidTimeDuration);
+            expectedResults = copyJsonObject(originalTime);
+            error = 'Bad Object';
+            statusCode = 400;
+            postBody = [
+            {
+                status: 400,
+                error: 'Bad object',
+                text: 'Field duration of time should be ' +
+                        'string but was sent as object'
```
tschuy: Should be number, I'd hope

tests/times.js
```diff
@@ -91,4 +91,624 @@ module.exports = function(expect, request, baseUrl) {
...
+            postObj = copyJsonObject(invalidTimeProject);
+            expectedResults = copyJsonObject(originalTime);
+            error = 'Bad Object';
+            statusCode = 400;
+            postBody = [
+            {
+                status: 400,
+                error: 'Bad object',
+                text: 'Field project of time should be ' +
+                        'string but was sent as object'
```
tschuy: For any of these 'invalid' ones, test a valid datatype with bad data -- here, a string that doesn't match the correct regex. Another test should be wirtten for bad datatype.

tests/times.js
ElijahCaine: The testing fixtures have ``"date_worked"`` set to ``null``, this reflects the state of the test data. iirc we haven't changed this to be a 'real' date.

tschuy: I think now might as well be when you change those. We really need them to be updated, and it doesn't look like we're doing it elsewhere.

tests/times.js
ElijahCaine: I will change the tests to do that, it's a bit more descriptive to set the object to ``{duration: patchedTime.duration}`` directly.

tests/times.js
ElijahCaine: On line 201 or 201 and 200?

tests/times.js
ElijahCaine: That's meant to be ``postBodies`` I changed this function half way through testing and forgot to change some parts. Thanks for catching that.

tests/times.js
ElijahCaine: Because it should be an array of possible bodies (it helps for making the tests generalized, I see if the body received was in the set of bodies I expect)

tests/times.js
ElijahCaine: It is ``postBodies`` in the general testing function. I figured I had to pass it something so I explicitly pass undefined and do the check later on. The difference in names is a bit confusing. I can change that.

tests/times.js
Kennric: I meant the 'created_at' - for data integrity, that should never be null

tschuy:
+1, these tests look good to implement.

Kennric:
I think they are good. +1

tests/times.js
```diff
@@ -957,4 +957,857 @@ module.exports = function(expect, request, baseUrl) {
...
+
+  describe('POST /times/:id', function() {
+    // The database's entry for `Whats Fresh`'s time entry
+    const originalTime = {
+      duration: 12,
+      user: 'tschuy',
+      project: 'wf',
+      notes: '',
+      activities: ['docs', 'dev'],
+      // jscs:disable
```
tschuy: can you get rid of all of the ``jscs:disable`` and ``jscs:enable`` bits? we use eslint now.

src/times.js
```diff
@@ -333,4 +333,220 @@ module.exports = function(app) {
...
+    passport.authenticate('local', function(autherr, user, info) {
+      if (!user) {
+        const err = errors.errorAuthenticationFailure(info.message);
+        return res.status(err.status).send(err);
+      }
+
+      const obj = req.body.object;
+
+      // Test duration value
+      if (obj.duration !== undefined && helpers.getType(obj.duration) === 'object') {
```
tschuy: too long

src/times.js
```diff
@@ -333,4 +333,220 @@ module.exports = function(app) {
...
+      }
+
+      // Test each activity
+      if (obj.activities !== undefined) {
+        /* eslint-disable prefer-const */
+        for (let activity of obj.activities) {
+          /* eslint-enable prefer-const */
+          if (helpers.getType(activity) !== 'string') {
+            const err = errors.errorBadObjectInvalidField('time', 'activities',
+            'slugs', 'array containing at least 1 ' + helpers.getType(activity));
```
tschuy: too long

src/times.js
```diff
@@ -333,4 +333,220 @@ module.exports = function(app) {
...
+            const err = errors.errorBadObjectInvalidField('time', 'activities',
+            'slugs', 'array containing at least 1 invalid slug');
+            return res.status(err.status).send(err);
+          }
+        }
+      }
+
+      // Test issue URI value
+      if (obj.issue_uri !== undefined && !validUrl.isWebUri(obj.issue_uri)) {
+        const err = errors.errorBadObjectInvalidField('time', 'issue_uri', 'URI',
```
tschuy: too long

src/times.js
```diff
@@ -333,4 +333,220 @@ module.exports = function(app) {
...
+
+      // Test notes value
+      if (obj.notes !== undefined && helpers.getType(obj.notes) !== 'string') {
+        const err = errors.errorBadObjectInvalidField('time', 'notes',
+        'string', helpers.getType(obj.notes));
+        return res.status(err.status).send(err);
+      }
+
+      // I think this isn't necessary but there's a test for it
+      if (obj.key !== undefined) {
```
tschuy: this is **not** how to check for arbitrary extra keys, this is how to check for the extra key ``key``. You'll want to, above,  have a list of valid fields (you could get that from the fields by mapping the fields to their names: ``const fieldNames = fields.map(function(field) { return field.name });``) and then iterate through the ``obj`` the user posted, and then if any of them aren't in ``fieldNames`` throw the error.

src/times.js
```diff
@@ -333,4 +333,220 @@ module.exports = function(app) {
...
+              'times.project as project', 'times.notes as notes',
+              'times.issue_uri as issue_uri',
+              'times.date_worked as date_worked',
+              'times.created_at as created_at',
+              'times.updated_at as updated_at', 'times.id as id',
+              'users.username as owner', 'projectslugs.name as projectName')
+          .where('times.id', '=', req.params.id).innerJoin('users', 'users.id',
+                  'times.user').innerJoin('projectslugs', 'projectslugs.id', 'times.project')
+          .then(function(time) {
+        if (user.username !== time[0].owner) {
```
tschuy: We actually need to update the tests to reflect the new permissions model, and then update this here too. Only the time owner, *or an admin* can modify times.

src/times.js
```diff
@@ -333,4 +333,220 @@ module.exports = function(app) {
...
+          const err = errors.errorAuthorizationFailure(user.username,
+            'create objects for ' + time[0].owner);
```
tschuy: update

src/times.js
```diff
@@ -333,4 +333,220 @@ module.exports = function(app) {
...
+          } else {
+            time[0].user = time[0].user;
+          }
+
+          const projectName = obj.project || time[0].projectName;
+          helpers.checkProject(projectName).then(function() {
+            knex('projectslugs').select('project').where('name', '=', obj.project).then(function(projectId) {
+              if (projectId[0] !== undefined) {
+                time[0].project = projectId[0].project;
+              } else {
```
tschuy: this else does nothing (assigning something back to itself).

When would projectId[0] be undefined?

src/times.js
```diff
@@ -333,4 +333,220 @@ module.exports = function(app) {
...
+        const username = obj.user || time[0].owner;
+        helpers.checkUser(username, username).then(function(userId) {
+          if (userId !== undefined) {
+            time[0].user = userId;
+          } else {
+            time[0].user = time[0].user;
+          }
+
+          const projectName = obj.project || time[0].projectName;
+          helpers.checkProject(projectName).then(function() {
```
tschuy: this helper resolves with the project's id as its first argument.

MaraJade:
@tschuy Fixed issues

tschuy:
Looks good to me. @ElijahCaine ?

Kennric:
Looks good to me +1

PR #114, opened 2015-08-11
rearrange before/after database hooks

tschuy:
This allows tests to fail and not break the next round of tests. Now, for every test, it will BEFORE THE TEST clear the database and reload the fixtures.

### DIFF
--- a/tests/test.js
+++ b/tests/test.js
@@ -49,11 +49,9 @@ describe('Endpoints', function() {
     });
 
     beforeEach(function(done) {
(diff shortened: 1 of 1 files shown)


LyonesGamer:
Can confirm it works. +1 Also PR #90 is blocked on this PR.

Kennric:
Makes sense to me, I am surprized the test runner doesn't handle this automatically. +1

tschuy:
Was a little surprised about that myself.

PR #115, opened 2015-08-12
POST /activities/<slug> Tests

### LINKED ISSUE #95: [13] PATCH Activities Endpoint
Create tests and implementation that allows the user to update/edit an existing activities object.  Users should be able to POST to /activity/:slug with a JSON body that only contains the part that is being updated.

athai:
Fixes #95.

### DIFF
--- a/src/activities.js
+++ b/src/activities.js
@@ -4,6 +4,7 @@ module.exports = function(app) {
   const knex = app.get('knex');
   const errors = require('./errors');
   const helpers = require('./helpers')(app);
+  const passport = require('passport');
 
   app.get(app.get('version') + '/activities', function(req, res) {
     knex('activities').then(function(activities) {
@@ -76,6 +77,45 @@ module.exports = function(app) {
         const err = errors.errorServerError(error);
         return res.status(err.status).send(err);
       });
+  });
+
+    app.post(app.get('version') + '/activities/:slug',
+    function(req, res, next) {
+        passport.authenticate('local', function(autherr, user, info) {
+            if (!user) {
+                let err = errors.errorAuthenticationFailure(info.message);
+                return res.status(err.status).send(err);
+            }
+
+            let obj = req.body.object;
+
+            let validKeys = ['name', 'slug']
+            for (let key in obj) {
+                if (validKeys.indexOf(key) === -1) {
+                    let err = errors.errorBadObjectUnknownField(
+                        'activity', key);
+                    return res.status(err.status).send(err);
+                }
+            }
+
+            let fields = [
+                {name: 'name', type: 'string', required: false},
+                {name: 'slug', type: 'string', required: false},
+            ];
+
+            // Rebase later
+            let validationFailure = helpers.validateFields(obj, fields);
+            if (validationFailure) {
+                let err = errors.errorBadObjectInvalidField(
+                    'activity',
+                    validationFailure.name,
+                    validationFailure.type,
+                    validationFailure.actualType
+                );
+                return res.status(err.status).send(err);
+            }
+        });
     });
   });
+
 };
--- a/src/activities.js
+++ b/src/activities.js
@@ -83,31 +83,35 @@ module.exports = function(app) {
   app.post(app.get('version') + '/activities/:slug', function(req, res, next) {
     passport.authenticate('local', function(autherr, user, info) {
       if (!user) {
-        let err = errors.errorAuthenticationFailure(info.message);
+        const err = errors.errorAuthenticationFailure(info.message);
         return res.status(err.status).send(err);
       }
 
-      let currObj = req.body.object;
+      const currObj = req.body.object;
 
-      let validKeys = ['name', 'slug'];
+      const validKeys = ['name', 'slug'];
+
+      /* eslint-disable prefer-const */
       for (let key in currObj) {
+      /* eslint-enable prefer-const */
+
       // indexOf return -1 if the parameter it not in the array
       // so this will return true in the slug/name DNE
         if (validKeys.indexOf(key) === -1) {
-          let err = errors.errorBadObjectUnknownField('activity', key);
+          const err = errors.errorBadObjectUnknownField('activity', key);
           return res.status(err.status).send(err);
         }
       }
 
-      let fields = [
+      const fields = [
         {name: 'name', type: 'string', required: false},
         {name: 'slug', type: 'string', required: false},
       ];
 
       // Rebase later to use this helper function
-      let validationFailure = helpers.validateFields(currObj, fields);
+      const validationFailure = helpers.validateFields(currObj, fields);
       if (validationFailure) {
-        let err = errors.errorBadObjectInvalidField(
+        const err = errors.errorBadObjectInvalidField(
           'activity',
           validationFailure.name,
           validationFailure.type,         // expected type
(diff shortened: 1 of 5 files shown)


tests/activities.js
```diff
@@ -83,4 +89,364 @@ module.exports = function(expect, request, baseUrl) {
...
+        // Attempt at patching
+        var patchedActivity = {
+            name:'TimeSync Documentation',
+            slug:['docs', 'dev-docs']
+        };
+
+        // Control that the updated activity is being compared to
+        var originalActivity =  {
+            name:'Documentation',
+            slug:'docs',
```
ElijahCaine: Should this be a list or a string? The ``patchedActivity`` slug is a list.

tests/activities.js
```diff
@@ -83,4 +89,364 @@ module.exports = function(expect, request, baseUrl) {
...
+        var patchedSlug = patchedActivity.slug;
+
+        // Bad object
+        var badActivity = {
+            name: '',
+            slug: ''
+        };
+
+        var badPatchedName = badActivity.name;
+        var badPatchedSlug = badActivity.slug;
```
ElijahCaine: This should probably look something more like:
```
var badPatchedName = {name: badActivity.name}
var badPatchedSlug = {slug: badActivity.slug}
```
or in the tests replace instances of ``badPatched<Var>`` with ``{<var>: badActivity.<var>}`` as it is a bit more descriptive what you are sending to the endpoint. I opted for this new way of doing things in parts of #110. The way it is now you are not sending an object just the value when you assign ``badPatched<Var>`` to ``requestOptions.form.object``

tests/activities.js
```diff
@@ -83,4 +89,364 @@ module.exports = function(expect, request, baseUrl) {
...
+        // Control that the updated activity is being compared to
+        var originalActivity =  {
+            name:'Documentation',
+            slug:'docs',
+            id: 1
+        };
+
+        // Assigning the fields of patchedActivity to variables
+        var patchedName = patchedActivity.name;
+        var patchedSlug = patchedActivity.slug;
```
ElijahCaine: This doesn't do what I think you want it to. I have a comment about the same code but for badPatchedName that addresses the issue.

tests/activities.js
```diff
@@ -83,4 +89,364 @@ module.exports = function(expect, request, baseUrl) {
...
+                    var jsonBody = JSON.parse(body);
+                    expect(jsonBody).to.deep.equal(expectedResult);
+                    done();
+                });
+            });
+        });
+
+        it('successfully updates the activity slug', function(done) {
+            requestOptions.form = postArg;
+            requestOptions.form.object = patchedSlug;
```
ElijahCaine: Same problem as with line 178

tests/activities.js
```diff
@@ -83,4 +89,364 @@ module.exports = function(expect, request, baseUrl) {
...
+            requestOptions.form = postArg;
+            requestOptions.form.object = badPatchedName;
+
+            request.post(requestOptions.url, function(err, res, body) {
+                var jsonBody = JSON.parse(body);
+                var expectedError = {
+                    status: 400,
+                    error: 'Bad object',
+                    text: 'Field name of activity should be a string but was' +
+                          ' received as an empty string'
```
ElijahCaine: Is the exact wording of the error supposed to be 'as an empty string' or 'as empty string'? I haven't found docs that specify one way or another, but we should stay consistent.

tests/activities.js
```diff
@@ -83,4 +89,364 @@ module.exports = function(expect, request, baseUrl) {
...
+                    status: 400,
+                    error: 'Bad object',
+                    text: 'Field slug of activity should be slugs but was ' +
+                          'received as non-slug strings'
+                };
+
+                expect(jsonBody).to.deep.equal(expectedError);
+                expect(res.statusCode).to.equal(400);
+
+                request.get(requestOptions.url, function(err, res, body) {
```
ElijahCaine: This ``get`` block is repeated almost verbatim a few times. We might want to consider writing it as a function that we call each time to verify the endpoint is unchanged.

ElijahCaine:
I listed a few comments and concerns. These tests look really good so far and are almost good to go. Nice work :smiley:

tests/activities.js
```diff
@@ -83,4 +89,421 @@ module.exports = function(expect, request, baseUrl) {
...
+            * Were they added correctly (not overwriting things you don't want
+              to overwrite, right?)
+            * Compare the post response to the get response
+            * Twice
+            * Check out pull 97 for a rough basis on how to maybe do things */
+
+        // Attempt at patching
+        var patchedActivity = {
+            name:'TimeSync Documentation',
+            slug:['dev-docs']
```
LyonesGamer: This should be a string. Activities can only have one slug.

tests/activities.js
```diff
@@ -83,4 +89,421 @@ module.exports = function(expect, request, baseUrl) {
...
+
+                var jsonBody = JSON.parse(body);
+
+                expect(jsonBody).to.deep.equal(originalActivity);
+                done();
+            });
+        }
+
+        it('successfully updates the activity', function(done) {
+            requestOptions.form = postArg;
```
LyonesGamer: I just checked the RequestJS docs, and it looks like we want to be using `body`, not `form` here. `form` sets it in querystring format, and sets the Content-Type to application/x-www-form-urlencoded, only for the fact that `json` is set to true to cause it to re-parse it back to JSON and re-set the Content-Type to application/json. Just use `body` to avoid that.

tests/activities.js
```diff
@@ -83,4 +89,421 @@ module.exports = function(expect, request, baseUrl) {
...
+            requestOptions.form.object = patchedActivity;
+
+            request.post(requestOptions.url, function(err, res, body) {
```
LyonesGamer: This should just be `request.post(requestOptions, ...)`.

tests/activities.js
```diff
@@ -83,4 +89,421 @@ module.exports = function(expect, request, baseUrl) {
...
+                    status: 400,
+                    error: 'Bad object',
+                    text: 'Field name of activity should be string but was' +
+                          ' received as empty string'
+                };
+
+                expect(jsonBody).to.deep.equal(expectedError);
+                expect(res.statusCode).to.equal(400);
+
+                /*request.get(requestOptions.url, function(err, res, body) {
```
LyonesGamer: Any reason this is commented out?

LyonesGamer:
I don't see anything outstanding or problematic. Linter passes. Only tests failing are the new ones. +1

tschuy:
+1, looks good. Any minor problems can be solved during implementation

src/activities.js
```diff
@@ -78,4 +79,92 @@ module.exports = function(app) {
...
+      // checks non-empty string was sent to update slug
+      if (currObj.slug !== undefined && currObj.slug.length === 0) {
+        const err = errors.errorBadObjectInvalidField(
+          'activity', 'slug', 'slug', 'empty string');
+        return res.status(err.status).send(err);
+      }
+
+      // checks for valid slugs
+      if (!helpers.validateSlug(req.params.slug)) {
+        const err = errors.errorInvalidIdentifier('slug', '!._cucco');
```
LyonesGamer: The second parameter should be the slug that was received.

athai: That is the slug that was received though...?

LyonesGamer: Will '!._cucco' always be the slug that was sent? By anyone who would ever use this endpoint?

LyonesGamer:
Looks good, tests and linter pass. +1

Kennric:
Looks good to me. +1

PR #116, opened 2015-08-13
Add delete failure to list of errors

athai:
Error to return upon a refusal of a delete request.

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -168,7 +168,7 @@ module.exports = {
     /*
      * Error 10: Request failure. Used when a user attempts to GET, POST,
      * DELETE, etc. an object that is not allowed.
-     * 
+     *
      * param objectType (string): The name of the supplied object type
     */
     errorRequestFailure: function(objectType) {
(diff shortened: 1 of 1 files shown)


tschuy:
``project is associated with time`` seems like a very opaque error. Perhaps ``project has associated times``?

tschuy:
Looks good. Vague, but I suppose we can't really do better if we want it to be generic. @Kennric

Kennric:
Maybe too vague? Ideally, we would like to say something like "Method DELETE not allowed for activity: coding has times associated.", but this might be complicated by using the same error for things caught at the express level vs the endpoint code level.

Also, at what point are we setting the requisite "Allow" header?

ElijahCaine:
This matches up with osuosl/timesync#20 and I agree with that. Let's make sure we merge osuosl/timesync#20 as well. (update the docs -> update the code)

tschuy:
I merged osuosl/timesync#20, so this is good to go whenever.

tschuy:
Looks good to me.

Kennric:
Matches the current API spec - at some point I'd liek to make this a bit more sophisticated, but this is good to go for now. +1

PR #117, opened 2015-08-13
Speed up tests with in-memory database

tschuy:
This also reduces the number of database connections from two plus one per test run to only one for the entire run.


Kennric:
Looks like a nice improvement on test performance, and the code is cleaner as well. +1

ElijahCaine:
```
  56 passing (1s)
```
+1

(code looks good, pretty clean too, mostly commented. I approve)

PR #119, opened 2015-08-13
set jscs to allow snake_case properties

tschuy:
This means we won't need to keep wrapping things in ``jscs:disable`` and ``jscs:enable``. Should have googled this before.


Kennric:
That helps readability a lot. +1

LyonesGamer:
In the title did you mean snake_case? It encourages camelCase at the moment.

tschuy:
Yeah :)

LyonesGamer:
Seems to work, jscs doesn't complain for snake_case properties. Simple change. +1

PR #120, opened 2015-08-14
Allow Istanbul to ignore lines and branches which we cannot test.

LyonesGamer:
Prevent branches like those which handle database errors from artificially decreasing our test coverage, allowing us to reasonably reach 100% coverage. Helpful because previously I'd seen ~80% coverage and figured it was fine, but going through this allowed me to see areas that are *actually* not covered but should be.

Note that ignores have only been added, and *should* only be added in the future, where tests ***cannot*** cover a piece of code. Examples include catch functions on database calls (because we cannot reliably test a failing database), multiple returns where the return which is called depends on timing (e.g. in src/projects.js), etc.

### DIFF
--- a/migrations/20150101000000_00.js
+++ b/migrations/20150101000000_00.js
@@ -35,6 +35,7 @@ exports.up = function(knex, Promise) {
     });
 };
 
+/* istanbul ignore next */
 exports.down = function(knex, Promise) {
     return knex.schema.dropTable('projects');
 };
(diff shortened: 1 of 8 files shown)


Kennric:
Interesting - I think making sure the right code is covered is more important than high coverage numbers, I'm for this in principle. Are there any possible downsides or gotchas? It does require another layer of education/judgment for future devs to know what can safely be ignored.

LyonesGamer:
Yeah, I mean, the only gotcha is a future dev seeing low coverage numbers on their branch and deciding to use ignores to cover it up rather than write more tests to fix the problem. But that's exactly the kind of reason why we have things like code reviews.

PR #134, opened 2015-08-19
AirBnB style audit

tschuy:
This'll include things like moving to ``let`` and ``const``, as well as across-the-board ``strict`` mode.


LyonesGamer:
I can't say I entirely followed everything, but I don't see anything that stood out to me and it seems to be a better, more consistent style. +1

Kennric:
I don't see any red flags, it looks like a pretty comprehensive tidying up. +1

Issue #129, opened 2015-08-18
[5] Review tests for query parameters


LyonesGamer:
Query parameters have tests written but they need to be reviewed and potentially compressed/refactored.

LyonesGamer:
Spent 7 hours on this on 8/19.

LyonesGamer:
Spent 6 hours on this on 8/20.

LyonesGamer:
Although the PR has not been merged yet, the tests have been reviewed, and therefore this issue is closed.

PR #135, opened 2015-08-19
Review and update tests for query string parameters

LyonesGamer:
Fixes issue #129


tests/times.js
```diff
@@ -5,16 +5,1155 @@ module.exports = function(expect, request, baseUrl) {
...
+        expect(getRes.statusCode).to.equal(200);
+        expect(jsonBody).to.deep.have.same.members(expectedResults);
+        done();
+      });
+    });
+  });
+
+  describe('GET /times?dateStart=:start', function() {
+    it('should return all times after a date', function(done) {
+      request.get(baseUrl + 'times?dateStart=2015-04-20',
```
tschuy: is dateStart the format we want to use for our query parameters?

LyonesGamer: `start` and `end` could be fine too if that's what you think. I just feel like `dateStart` and `dateEnd` clarifies that we're talking about a start and end point in *time*. As opposed to....I don't know. ID?

tschuy:
These look good to me, but there's a lot of stuff here. Cautious +1.

tests/times.js
tschuy: @Kennric what do you think? Personally I think either ``snake_case`` or just ``start`` and ``end``.

Kennric:
We are trying to avoid "should" in the it() statements.

I am not seeing tests for invalid params - sending garbage or empty values

What do we expect when we send a param we don't handle? i.e. /times/?this_is_bad=27

tests/times.js
Kennric: 'start' and 'end' are a little cleaner, I think it is implicit that these should be dates

tschuy:
APIs I've used ignore extra fields in query parameters.

LyonesGamer:
Agreed. Unlike path params, query params tend to be just dealt with. Ones that shouldn't exist are ignored, ones with bad values are ignored, ones that repeat too many times just use the first value(s), ones that should exist but don't assume a default parameter.

Kennric:
I am fine with ignoring unhandled params, but I think if we have a handled one with a bad value, an error response would be useful for debugging (did I send the date in right format?)

Also, for sanity on the UI end, knowing that we'll get an error for, say, an activity that doesn't exist, would be very nice - I don't want a list of times I -assume- are for a particular activity, but really they are for all activities because my activity slug was mis-spelled, or I sent an integer id instead of a slug by mistake.

I believe in the What's Fresh api, we could send an error message back along with partial or unfiltered data - in theory we could say something like "here is the data, but it's not filtered by your parameter because your parameter value wasn't found" which could be handy for displaying data plus a warning message.

tschuy:
I would tend to agree that we should definitely error when we get invalid things passed to valid parameters

tests/times.js
```diff
@@ -5,16 +5,1269 @@ module.exports = function(expect, request, baseUrl) {
...
+      });
+    });
+
+    it('returns an error for a non-existent user', function(done) {
+      request.get(baseUrl + 'times?user=fakeuser', function(getErr, getRes, getBody) {
+        const jsonBody = JSON.parse(getBody);
+        const expectedResults = {
+          status: 400,
+          error: 'Bad Query Value',
+          text: 'Parameter user contained invalid value "fakeuser"',
```
tschuy: Will this be able to distinguish between invalid user ``#$&##$#%$#""}>].]`` and invalid user ``bobtypodhisusernameagain``?

LyonesGamer: I don't see why it wouldn't? But I can add another test if you want.

Kennric:
I think this is looking pretty good, but please get another review. +1

tests/times.js
```diff
@@ -5,16 +5,1269 @@ module.exports = function(expect, request, baseUrl) {
...
+        expect(getRes.statusCode).to.equal(200);
+        expect(jsonBody).to.deep.have.same.members(expectedResults);
+        done();
+      });
+    });
+  });
+
+  describe('GET /times?user=:user&startDate=:start', function() {
+    it('returns all times for a user after a date', function(done) {
+      request.get(baseUrl + 'times?user=deanj&startDate=2015-04-20',
```
tschuy: start

tests/times.js
```diff
@@ -5,16 +5,1269 @@ module.exports = function(expect, request, baseUrl) {
...
+
+        expect(getErr).to.equal(null);
+        expect(getRes.statusCode).to.equal(200);
+        expect(jsonBody).to.deep.have.same.members(expectedResults);
+        done();
+      });
+    });
+  });
+
+  describe('GET /times?user=:user&endDate=:end', function() {
```
tschuy: end and two lines down

tests/times.js
tschuy: what will the error be for a non-slug be?

LyonesGamer: ``Parameter user contained invalid value "#$&##$#%$#""}>].]"``

Unless you think we really need to split it into two errors.

tschuy: We split it into two everywhere else *shrug* @Kennric ?

tschuy:
+1, looks good. Will start implementing mine tomorrow.

LyonesGamer:
Rebased with develop.

iankronquist:
This PR hasn't been touched in a month. Presumably this is a feature we still want which has not yet been implemented. Does this PR need to be adopted?

LyonesGamer:
All feature branches have been merged in now, one last chance for a holistic review before merge into develop.

iankronquist:
It looks like this needs a rebase?

LyonesGamer:
All tests pass, all functionality works as expected.

Kennric:
+1

MaraJade:
+1

PR #136, opened 2015-08-20
add bad datatype checking to POST projects/

tschuy:
Currently if you POST a string to the slug field, it'll crash the server when it attempts to run a ``.filter`` on it. Check your datatypes, folks!

### DIFF
--- a/src/projects.js
+++ b/src/projects.js
@@ -152,11 +152,17 @@ module.exports = function(app) {
       const obj = req.body.object;
 
       // run various checks
-
-      // check validity of uri syntax
-      if (obj.uri && !validUrl.isWebUri(obj.uri)) {
-        const err = errors.errorInvalidIdentifier('uri', obj.uri);
-        return res.status(err.status).send(err);
+      // valid keys
+      const validKeys = ['name', 'uri', 'owner', 'slugs'];
+      /* eslint-disable prefer-const */
+      for (let key in obj) {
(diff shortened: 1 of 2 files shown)


tschuy:
@Kennric

tschuy:
Fixes #137.

Kennric:
This seems to address the issue, and I don't see any problems. It should be noted that the tests check by sending a single wrong type to the endpoint, so it isn't a comprehensive test for all possible bad types, but verifies that a single wrong type is caught. I don't think that's a problem at this point, based on how the code is doing the check. +1

src/projects.js
```diff
@@ -171,6 +177,30 @@ module.exports = function(app) {
...
+      if (validationFailure) {
+        const err = errors.errorBadObjectInvalidField('project',
+          validationFailure.name, validationFailure.type,
+          validationFailure.actualType);
+        return res.status(err.status).send(err);
+      }
+
+      // check validity of uri syntax
+      if (obj.uri && !validUrl.isWebUri(obj.uri)) {
+        const err = errors.errorInvalidIdentifier('uri', obj.uri);
```
LyonesGamer: This needs to be changed to a Bad Object error (type Invalid Field).

LyonesGamer:
+1

Kennric:
Good catch on the object errors @LyonesGamer  +1

Issue #131, opened 2015-08-18
[5] Date range parameter on time requests


LyonesGamer:
When making a GET request to /times, one should be able to request with ?dateStart= and ?dateEnd= to get times with a date_worked within a certain date range. Dates will be in ISO 8601 format.

LyonesGamer:
Spent 4 hours on 8/24.

LyonesGamer:
Spent 3 hours on 8/28, 2 hours on 8/31.

LyonesGamer:
Fixed by #142

PR #142, opened 2015-08-24
Select times within a given date range

LyonesGamer:
Allow a user to select times only within a given date range, using start and end query parameters. Fixes issue #131.

### DIFF
--- a/src/times.js
+++ b/src/times.js
@@ -104,7 +104,7 @@ module.exports = function(app) {
               return res.status(err.status).send(err);
             }
 
-            timesQ = timesQ.andWhere('date_worked', '>', startDate.getTime());
+            timesQ = timesQ.andWhere('date_worked', '>=', startDate.getTime());
           }
 
           if (req.query.end) {
@@ -134,18 +134,14 @@ module.exports = function(app) {
               }
             }
 
-            timesQ = timesQ.andWhere('date_worked', '<', endDate.getTime());
+            timesQ = timesQ.andWhere('date_worked', '<=', endDate.getTime());
           }
-
-          console.log(timesQ.toString());
 
           let activitiesDone = false;
           let projectsDone = false;
           let usersDone = false;
 
           timesQ.then(function(times) {
-            console.log(times.length);
-
             if (times.length === 0) {
               return res.send([]);
             }
--- a/src/times.js
+++ b/src/times.js
@@ -86,6 +86,57 @@ module.exports = function(app) {
             }));
           }
 
+          if (req.query.start) {
+            let start;
+            if (typeof req.query.start === 'string') {
+              start = req.query.start;
+            } else if (helpers.getType(req.query.start) === 'array') {
+              start = req.query.start[0];
+            } else {
+              const err = errors.errorBadQueryValue('start', req.query.start);
+              return res.status(err.status).send(err);
+            }
+
+            const startInt = Date.parse(start);
+
+            if (!startInt || startInt > Date.now()) {
+              const err = errors.errorBadQueryValue('start', req.query.start);
+              return res.status(err.status).send(err);
+            }
+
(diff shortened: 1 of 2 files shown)


src/times.js
```diff
@@ -8,7 +8,60 @@ module.exports = function(app) {
...
+      }
+
+      const startInt = Date.parse(start);
+
+      if (!startInt || startInt > Date.now()) {
+        const err = errors.errorBadQueryValue('start', req.query.start);
+        return res.status(err.status).send(err);
+      }
+
+      query = query.andWhere(knex.raw('date_worked >= date("' + start + '")'));
```
tschuy: you should comment this

src/times.js
```diff
@@ -8,7 +8,67 @@ module.exports = function(app) {
...
+      const startInt = Date.parse(start);
+
+      if (!startInt || startInt > Date.now()) { // Disallow future dates
+        const err = errors.errorBadQueryValue('start', req.query.start);
+        return res.status(err.status).send(err);
+      }
+
+      // Date comparison in SQLite uses string comparison (which works on ISO dates)
+      // We need to use a raw query to gain access to the date() function in SQL
+      query = query.andWhere(knex.raw('date_worked >= date("' + start + '")'));
```
tschuy: Will this work cross-database?

src/times.js
```diff
@@ -8,7 +8,67 @@ module.exports = function(app) {
...
+    }
+
+    // Identical to the above
```
tschuy: if this is identical, would factoring out be appropriate?

src/times.js
LyonesGamer: No, and that's a difficulty that SQLite causes us by choosing not to natively have datetime functionality. This will need to be changed if we swap databases. I can try to figure out something to work cross-database, but it will likely be hacky.

src/times.js
LyonesGamer: My mistake, that comment is poorly worded. I meant "works in the same way". It's not identical because they're two separate params.

It may be possible to refactor it to be shorter and potentially faster, but I doubt it'd be any less than messy and difficult to read. It may be worth a try though.

src/times.js
tschuy: ok. We should figure out what kind of db we'll be hosting this on. Can you make an issue about making timesync multi db compatible?

Kennric:
Lets put this aside until we can develop with mysql/postrges - I'd rather not pin our code to our dev database.

LyonesGamer:
Blocked on PR #151

LyonesGamer:
Rebased to changes in parent branch.

tschuy:
correct tests pass. let's check out the code.

src/times.js
```diff
@@ -86,6 +86,57 @@ module.exports = function(app) {
             }));
           }
 
+          if (req.query.start) {
+            let start;
+            if (typeof req.query.start === 'string') {
+              start = req.query.start;
+            } else if (helpers.getType(req.query.start) === 'array') {
+              start = req.query.start[0];
```
tschuy: check type

src/times.js
```diff
@@ -86,6 +86,57 @@ module.exports = function(app) {
             }));
           }
 
+          if (req.query.start) {
```
tschuy: why did you decide to not factor the processing out?

tests/times.js
```diff
@@ -59,7 +59,7 @@ module.exports = function(expect, request, baseUrl) {
 
         expect(getErr).to.be.a('null');
         expect(getRes.statusCode).to.equal(200);
-        expect(JSON.parse(getBody)).to.deep.have.same.members(expectedResults);
+        expect(JSON.parse(getBody)).to.deep.equal(expectedResults);
```
tschuy: fix these

tschuy:
+1

Kennric:
+1

PR #143, opened 2015-08-24
Update dependencies

iankronquist:
I can see that @tschuy's update to the airbnb style guide made one of the version numbers bump. Also, passport got a new minor version. All tests pass.


tschuy:
+1. @Kennric @athai @MaraJade ?

LyonesGamer:
If updating these didn't break anything (and it looks like they didn't), then I say +1.

PR #151, opened 2015-08-27
Changed code to use integer representations of dates internally.

LyonesGamer:
The code now stores dates as bigint, a signed 64-bit integer, in the database. However, the API still uses ISO format dates.


tschuy:
Tests, think of the tests! (They're failing.)

migrations/20150101000000_00.js
```diff
@@ -13,7 +13,7 @@ exports.up = function(knex) {
     table.integer('project').references('id').inTable('projects').notNullable();
     table.string('notes');
     table.string('issue_uri');
-    table.timestamp('date_worked');
+    table.bigInteger('date_worked');
     table.timestamps();
```
tschuy: make sure this makes the right things

tests/fixtures/test_data.json
```diff
@@ -15,6 +15,11 @@
       "name":"Systems",
       "slug":"sys",
       "id": 3
+    },
+    {
+      "name":"Meetings",
+      "slug":"meeting",
+      "id":4
     }
```
tschuy: this makes me wince

tests/times.js
```diff
@@ -5,26 +5,1287 @@ module.exports = function(expect, request, baseUrl) {
   what should be returned */
   describe('GET /times', function() {
     it('returns all times in the database', function(done) {
-      request.get(baseUrl + 'times', function(err, res, body) {
+      request.get(baseUrl + 'times', function(getErr, getRes, getBody) {
```
tschuy: but not as much as this does

migrations/20150101000000_00.js
```diff
@@ -13,8 +13,9 @@ exports.up = function(knex) {
     table.integer('project').references('id').inTable('projects').notNullable();
     table.string('notes');
     table.string('issue_uri');
-    table.timestamp('date_worked');
-    table.timestamps();
+    table.bigInteger('date_worked').notNullable();
+    table.bigInteger('created_at').notNullable();
+    table.bigInteger('updated_at').defaultTo(-1);
```
tschuy: why not null?

LyonesGamer:
Rebased to new rebase in parent branch.

src/times.js
```diff
@@ -8,131 +8,206 @@ module.exports = function(app) {
   const passport = require('passport');
 
   app.get(app.get('version') + '/times', function(req, res) {
-    knex('times').then(function(times) {
-      if (times.length === 0) {
-        return res.send([]);
-      }
+    let activitiesList = req.query.activity;
```
tschuy: why is this a diff?

tests/times.js
```diff
@@ -5,26 +5,1287 @@ module.exports = function(expect, request, baseUrl) {
...
+
+        expect(getErr).to.equal(null);
+        expect(getRes.statusCode).to.equal(200);
+        expect(JSON.parse(getBody)).to.deep.equal(expectedResults);
+        done();
+      });
+    });
+  });
+
+  describe('GET /times?user=:user', function() {
```
tschuy: why are there new tests in this commit?

migrations/20150101000000_00.js
LyonesGamer: We've always agreed that they shouldn't be null. It just wasn't reflected in the schema.

tschuy:
+1

tests/times.js
```diff
@@ -59,7 +59,7 @@ module.exports = function(expect, request, baseUrl) {
 
         expect(getErr).to.equal(null);
         expect(getRes.statusCode).to.equal(200);
-        expect(JSON.parse(getBody)).to.deep.have.same.members(expectedResults);
+        expect(JSON.parse(getBody)).to.deep.equal(expectedResults);
```
athai: Change ``.to.deep.equal`` back to ``.to.deep.have.same.members`` every time you're comparing the contents of an array. I think it looks good other than that.

tschuy:
+1

tschuy:
+1

athai:
+1

PR #152, opened 2015-08-27
Thai/schema change

athai:
Added a 'user roles' table with boolean values to determine which permissions a project's user has:
- manager
- member
- spectator (instead of data viewer but we should definitively decide on a name soon)

New users have been created as well.

### DIFF
--- a/tests/fixtures/README.md
+++ b/tests/fixtures/README.md
@@ -3,10 +3,28 @@ Fixtures
 
 ``test_data.json``
 ------------------
+Each user listed below has a different set of permissions for their respective
+projects.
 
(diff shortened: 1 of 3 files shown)


tests/fixtures/README.md
```diff
@@ -6,3 +6,7 @@ Fixtures
 
 1) deanj's password is ``pass``
 2) tschuy's password is ``password``
+3) patcht's password is ``drowssap``
```
tschuy: can you document people's permissions?

LyonesGamer:
Looks good to me. +1

tschuy:
+1

PR #154, opened 2015-08-27
add production database environment

tschuy:
This required updating the fixtures as well, since many of them took
advantage of sqlite being lazy and allowing bad foreign keys.

It also moves everything over to the NODE_ENV variable, which is the
"blessed" way of doing this kind of config.


tschuy:
the environment variable ``PG_CONNECTION_STRING`` should be set to a string like ``postgres://username:password@server:port/table_name``.

LyonesGamer:
Add some small documentation regarding PG_CONNECTION_STRING and it should be good.

LyonesGamer:
+1

Kennric:
I don't see any issues. +1

PR #164, opened 2015-09-10
add clear-database command

tschuy:
Once we get a testing env up for docker, I'll redo this a little to clear postgres as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "start": "node --harmony ./src/app.js",
     "devel": "nodemon --harmony ./src/app.js",
     "migrations": "NODE_ENV=mocha knex migrate:latest",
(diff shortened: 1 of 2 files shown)


Kennric:
Not the most elegant thing in the world, but appears to do the job. +1

iankronquist:
Since we are using transactions in tests and no longer using the Great Pyramid of TimeSync this code will not be duplicated and it has my +1.

tschuy:
I've updated this. It now doesn't set the indexes, since that's database-specific and not really needed for a dev env. It does however output whenever there's an error what exactly that error is!

LyonesGamer:
Looks good to me. +1

tschuy:
```
10:31:09   tschuy | kennric: what about #164>
10:33:48  kennric | tschuy: yes
```

PR #165, opened 2015-09-11
add ldap support

tschuy:
This PR is for discussion of LDAP login and its implementation. Currently, there are a few critical things:

~~1. We have no way to automatically test this. I'm not putting my ldap password in the tests and neither should you!~~
3. It increases setup complexity about 10x over by introducing four new env variables, at least one of which MUST be set. 

Should we change ``TIMESYNC_LOCAL`` to something more like ``TIMESYNC_DISABLE_LOCAL``, and make password auth default?

~~I'd also like to refactor my use of passport into something like ``app.authenticatedPost(`` that handles all of the authentication nonsense itself.~~


ElijahCaine:
After doing the ``npm create-account`` thing and running the following:
```
 (tschuy/passport-ldap) neon (timesync-node) 11:27 $ curl -v -H "Content-Type: application/json" -X POST -d '
    {
        "auth": {
            "type": "ldap",
            "username": "my_username",
            "password": "my_password"
        },
        "object": {
            "name": "nodejs-webapp",
            "owner": "my_username",
            "slugs": ["nodejs-webapp"]
        }
    }' http://localhost:8000/v1/projects/
```
*(Censored for reasons)*
I got the response:
```
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> POST /v1/projects/ HTTP/1.1
> User-Agent: curl/7.38.0
> Host: localhost:8000
> Accept: */*
> Content-Type: application/json
> Content-Length: 276
> 
* upload completely sent off: 276 out of 276 bytes
< HTTP/1.1 200 OK
< X-Powered-By: Express
< Content-Type: text/html; charset=utf-8
< Content-Length: 74
< ETag: W/"XX-XXXXXXXXXXXXXXXXXXXXXX"
< Date: Fri, 11 Sep 2015 18:25:48 GMT
< Connection: keep-alive
< 
* Connection #0 to host localhost left intact
{"name":"nodejs-webapp","owner":"my_username","slugs":["nodejs-webapp"],"id":1} 
```
Travis says the tests pass, my local box agrees (when I remember to set ``TIMESYNC_LOCAL=1``). I looked over the code a few times, everything looks good to me, although a few more comments would be appreciated. Other than that :+1: from me.

**TLDR:** Unit tests pass, manual tests pass, everything looks good.

ElijahCaine:
@tschuy For testing would it be possible to set env variables within the tests instead of putting that on the user? I've done that with Python, it should be possible in Node, it's just a matter of if it's worth it.

iankronquist:
@ElijahCaine we set env variables when running tests a lot already, see here:
https://github.com/osuosl/timesync-node/blob/develop/package.json#L16
I think he means set up for deploying the code, or wheeling it out for a spin in development. I could be wrong though.

README.md
```diff
@@ -115,3 +115,18 @@ $ cd timesync-api
...
+Authentication
+--------------
+
+To use password-based authentication, set the environment variable
+``TIMESYNC_LOCAL`` to 1.
+
+To use LDAP authentication, there are three environment variables that need to
+be set:
+
+1. ``TIMESYNC_LDAP``: set this to 1 to enable LDAP
```
iankronquist: What happens when you set  both `TIMESYNC_LOCAL` and `TIMESYNC_LDAP` to 1?
Why not make a `TIMESYNC_AUTH` variable which can have the values `ldap` or `local`? This allows us to add future auth methods without adding more environment variables.

package.json
```diff
@@ -40,7 +40,9 @@
     "eslint-plugin-react": "^3.2.3",
     "express": "^4.12.3",
     "knex": "^0.8.6",
+    "ldapjs": "mcavage/node-ldapjs",
```
iankronquist: Should this be under `optionalDependencies`?

tschuy: It's just fine in here. Things in optionalDependencies will still be installed, but will not stop npm if they fail to install. Since it'll install anyway, and if it's failing you have a problem (these are pure-javascript packages), I think dependencies is the place to put it.

README.md
tschuy: Then you get to use both ``TIMESYNC_LOCAL`` and ``TIMESYNC_LDAP``.

tschuy: do we want to force only one kind of connection? We can easily do that in this schema by only enabling one, but this gives us greater flexibility.

tschuy:
Crazy awful idea: we could do file permissions do -- ``local == 1``, ``ldap == 2``, ``google == 4``, etc. Then we can set ``TIMESYNC_AUTH_MODULES`` to ``1``, ``3``, ``4``, ``7``, etc along with their relevant other vars!

ElijahCaine:
@tschuy how about no

iankronquist:
I like the json permissions. I'd like to ask some people's permission in `#osu-lug`.
I've also seen this approach, but I kind of like the json array better:
`qemu -drive file=file,index=0,media=disk`
Please see [man qemu-kvm](http://linux.die.net/man/1/qemu-kvm) for an example.

LyonesGamer:
I agree with the JSON permissions, and overall I like the way this is setup. The code seems to work. I'm not familiar enough with LDAP to understand tests/login/ldapserver.js, but aside from that file everything looks good. Tentative +1, but someone who knows LDAP better should look at that file.

Kennric:
This looks pretty good, but I do like the idea of having a default authentication module set, such that everything works as easily as possible out of the box, and layers of complexity can be added with more configuration.

ElijahCaine:
@Kennric I'm assuming we would use password auth for 'Out of the Box' functionality. Is that reasonable?

Kennric:
@ElijahCaine yep

Kennric:
Looks pretty good to me. +1

LyonesGamer:
Looks good to me. +1

Issue #150, opened 2015-08-25
[3] Docs: Setting up a local instance of a docker db


athai:

LyonesGamer:
Superseded by #172

PR #166, opened 2015-09-11
Refactor code to work with postgres

athai:
WIP #150, #177. 

Note: There are knex.raw sql statements in the source code for posting projects, activities, and times.  They're acting as a temporary fix for an insertion error (where something is causing new objects to be inserted at id 1 instead of the next available id).

### DIFF
--- a/tests/projects.js
+++ b/tests/projects.js
@@ -785,8 +785,8 @@ module.exports = function(expect, request, baseUrl) {
         const expectedError = {
           status: 409,
           error: 'The slug provided already exists',
-          text: 'slugs ganeti-webmgr, gwm already exist',
-          values: ['ganeti-webmgr', 'gwm'],
+          text: 'slugs gwm, ganeti-webmgr already exist',
+          values: ['gwm', 'ganeti-webmgr'],
         };
 
         expect(body).to.deep.equal(expectedError);
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,6 +2,7 @@
   "extends": "airbnb",
   "rules": {
     // rules here take precedence over airbnb's
+    "jsx-quotes": 0,
     "func-names": 0,
     "strict": [2, "global"],
     "max-len": [2, 80, 4]
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,6 @@
   "extends": "airbnb",
   "rules": {
     // rules here take precedence over airbnb's
-    "jsx-quotes": 0,
     "func-names": 0,
     "strict": [2, "global"],
     "max-len": [2, 80, 4]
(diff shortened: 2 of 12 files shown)


tests/projects.js
```diff
@@ -752,8 +752,8 @@ module.exports = function(expect, request, baseUrl) {
         const expectedError = {
           status: 409,
           error: 'The slug provided already exists',
-          text: 'slugs ganeti-webmgr, gwm already exist',
-          values: ['ganeti-webmgr', 'gwm'],
+          text: 'slugs gwm, ganeti-webmgr already exist',
+          values: ['gwm', 'ganeti-webmgr'],
```
LyonesGamer: If you sort expectedError.values and body.values, it won't matter the order. You could also use include.members() to make the order not matter in the text. As an example, see https://github.com/osuosl/timesync-node/blob/tristan/query-string-tests/tests/times.js#L482

athai:
Using a postgres database has somehow broken the way insertion works. Instead of using the next available id, insertion is continually happening at id 1, resulting in this error:
```javascript
"error": "Server error"
"status": 500
"text": {
    "code": "23505"
    "detail": "Key (id)=(1) already exists."
     "file": "nbtinsert.c"
     "length": 146
     "line": "392"
     "name": "error"
     "routine": "_bt_check_unique"
     "severity": "ERROR"
}
```
This occurs every time an otherwise legitimate attempt is made to add a new entry to a table. The current workaround is including a knex.raw statement before every insertion in the source code like so:
```javascript
knex.raw("SELECT setval('times_id_seq', (SELECT MAX(id) FROM times))").then(function() {
  //insertion code
}
``` 
But that's not a solid long term solution. I can't figure out what's breaking the code though. Does anyone have any ideas as to what might be happening?

iankronquist:
I dumped the database schema from PostGres and it looks like the schema is correct and the ids have SEQUENCE set on them.
https://pastebin.osuosl.org/33886/

Additionally, inserting values into the table directly with sql statements works as expected. The next step I can see would be to log all sql statements run, see which ones fail, and run them manually if they look correct.
http://stackoverflow.com/questions/722221/how-to-log-postgresql-queries

iankronquist:
More fun: I tried to reproduce the failures on OS X with the same docker container, but [everything (mostly) worked](https://pastebin.osuosl.org/33916/). 
I then tried again on green, and got the same errors @athai was reporting.

iankronquist:
I found out why. On my mac I'm using Nodejs 4. Using Nodejs 4 on the workstations fixes the problems @athai reported. I don't know why, but I can tell you it works.

tschuy:
Fixed by #191

PR #167, opened 2015-09-14
use transactions for endpoints with multiple INSERT/DEL statements

### LINKED ISSUE #153: Use transactions
c'est bon pour la santé!

Right now if we suffer a bug, something that makes something throw, etc., our data is happily saved into some random incomplete state.

Using transactions would save us from this HORRIBLE, HORRIBLE fate.

tschuy:
Fixes #153 

This is pretty dang important. Oops.

@LyonesGamer @Kennric 

I'm open for ideas on how to test this -- we can't just insert failure state into our code, unfortunately, unless any of you know a good library for that kind of thing.

### DIFF
--- a/src/projects.js
+++ b/src/projects.js
@@ -237,18 +237,21 @@ module.exports = function(app) {
             name: obj.name,
           };
 
-          knex('projects').insert(insertion).then(function(projects) {
-            // project is a list containing the ID of the
-            // newly created project
-            const project = projects[0];
-            const projectSlugs = obj.slugs.map(function(slug) {
-              return {name: slug, project: project};
-            });
+          knex.transaction(function(trx) {
+            trx('projects').insert(insertion).then(function(projects) {
+              // project is a list containing the ID of the
+              // newly created project
+              const project = projects[0];
+              const projectSlugs = obj.slugs.map(function(slug) {
+                return {name: slug, project: project};
+              });
 
-            knex('projectslugs').insert(projectSlugs).then(function() {
-              obj.id = project;
-              res.send(JSON.stringify(obj));
-            });
+              return trx('projectslugs').insert(projectSlugs).then(function() {
+                obj.id = project;
+                trx.commit();
+                res.send(JSON.stringify(obj));
+              });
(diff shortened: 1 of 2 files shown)


src/projects.js
```diff
@@ -237,18 +237,28 @@ module.exports = function(app) {
...
-              obj.id = project;
-              res.send(JSON.stringify(obj));
-            });
+              return trx('projectslugs').insert(projectSlugs).then(function() {
+                obj.id = project;
+                res.send(JSON.stringify(obj));
+              });
+            }).then(trx.commit)
+            .catch(trx.rollback);
+          }).then(function() {
```
tschuy: what

ElijahCaine:
Had a good meetspace discussion with @tschuy about this code. It all seems reasonable and it gets the job done.

As far as testing goes this I agree that it would be very difficult to test due to all of the checks we currently do on inputs before any transactions *can* occur. The fact that the current tests work is about as thorough as we can get until we / if we find a library to induce errors in random parts of the code that would trigger a rollback.

I'm gonna give this a +1 :dizzy:

src/projects.js
```diff
@@ -237,18 +237,21 @@ module.exports = function(app) {
...
-            knex('projectslugs').insert(projectSlugs).then(function() {
-              obj.id = project;
-              res.send(JSON.stringify(obj));
-            });
+              return trx('projectslugs').insert(projectSlugs).then(function() {
+                obj.id = project;
+                trx.commit();
+                res.send(JSON.stringify(obj));
+              });
+            }).catch(trx.rollback);
```
LyonesGamer: Do we not need to send some sort of error back here?

iankronquist:
Can you mock out the checks we do? If we have some function checkInputIsOkay and we just make it return true, then we may be able to coerce our code into entering the failure case in the transaction. Just a thought.

tschuy:
I like that idea.I can mock out ``helpers.js``, I'll get started on this either this afternoon after a meeting or  tomorrow, depending on what we're doing with standupbot.

tschuy:
Superceded by #192

Issue #173, opened 2015-09-15
[5] Soft Delete Times + Activities + Projects & Update Select Statements & Hard Delete Slugs


ElijahCaine:
Update database statements to soft delete everything except slugs (to avoid slug conflict down the road).

**Soft Deleted:**
- Activities
- Projects
- Times

**Hard Deleted:**
- Project slugs

LyonesGamer:
From previous issue #163:

Before a project can be deleted, the following must be true:

the user deleting the project must have admin rights
slugs associated with the project must be deleted
userroles associated with the project must be deleted

ElijahCaine:
This has been closed in favor of three individual issues: #186 #187 and #188

PR #176, opened 2015-09-17
Soft Deletes

ElijahCaine:
Added `deleted_at` and `parent` fields to test fixtures.
Modified activitiy tests to expect a new object instead of an updated
object.
refs #173

### DIFF
--- a/tests/activities.js
+++ b/tests/activities.js
@@ -21,6 +21,8 @@ module.exports = function(expect, request, baseUrl) {
             updated_at: null,
             parent: null,
             id: 1,
+            delete_at: null,
+            parent: null,
           },
           {
             name: 'Development',
@@ -29,6 +31,8 @@ module.exports = function(expect, request, baseUrl) {
             updated_at: null,
             parent: null,
             id: 2,
+            delete_at: null,
+            parent: null,
           },
           {
             name: 'Systems',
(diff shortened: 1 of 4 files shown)


ElijahCaine:
**Status:**
The PR is still ready for review, it's mostly changes to the tests and adds a few more (DELETE tests for times). Most of the tests fail because anything involving patching (POST), DELETing, or even GETting (because we are adding two fields to every object) fails.

We also get a ``113) "before each" hook`` failure about 135 tests in (out of about 170 tests :disappointed:) so that's a problem I'm not sure how to solve. 

So That's A Thing

For reviewing I would look through *why* each test failed to see what changes I made, many of them are pretty reasonable (adding expected fields to objects for GET requests, some other changes for the POST and DELETE endpoints).

We have also been having many discussions about if the ``parent`` and ``deleted_at`` field are the right way to go about this (i.e., is this good for the API or just our implementation; try to make the right API decision not just the right implementation decision). For updates about the API changes checkout osuosl/timesync/pull/28.

ElijahCaine:
Because our plans have changed so much (and the rebase is fairly dicey between this and develop) I am closing this pull request in favor of a new one from scratch off of develop.

ElijahCaine:
Because our plans have changed so much (and the rebase is fairly dicey between this and develop) I am closing this pull request in favor of a new one from scratch off of develop.

PR #179, opened 2015-09-21
use babel with es7 async/await

tschuy:
This is what the code would look like with ``async/await``. It's beautiful, and it's way, **way** easier to read. (Seriously! Compare https://github.com/osuosl/timesync-node/blob/tschuy/passport-ldap/src/activities.js#L43 to https://github.com/osuosl/timesync-node/blob/tschuy/asyncawait/src/activities.js#L44)

This has some upsides:

* uses *the future*
* much easier to read
* we're using the trendy js of the week, Babel
* it's all still Javascript, just new Javascript

This has some downsides:

* you need to compile the code to run it without Babel...
* ... which means we're using a transpiler, so you might just wonder why we're not using Coffeescript or Typescript.

And here's the big one:

* ``async/await`` is [Stage 2 of TC39's standardization process](https://babeljs.io/docs/usage/experimental/), which is ``Draft`` stage. It'll probably get included, but there *might* be changes to the spec. We're not doing anything crazy with it, though, so it shouldn't affect many people.

And here's another big one:

* ``jshint`` doesn't support it yet (https://github.com/jshint/jshint/issues/2601). I'm considering working on this, though.

Over all, I really want to use this, and there's only a very small voice at the back of my head telling me no. Many other people/groups use await/async in production, though, and it's **scheduled to be moved into Candidate stage soon** (like September soon)!

Thoughts?


tschuy:
there are forks of JSHint we can use: https://github.com/appium/appium-es7-jshint

LyonesGamer:
I'm moving this to an issue instead of a PR. Async/await is still in draft stage, and I would like to wait for the API to mature a little more before we do this, plus the code in this PR is so out of date that it's largely useless to us.

Issue #169, opened 2015-09-15
Design how to integrate with github hooks.


lucywyman:

lucywyman:
Addressed by #182

LyonesGamer:
Addressed by https://github.com/osuosl/gitsync .

PR #182, opened 2015-09-21
Proposal for github webhooks

lucywyman:
This PR is intended to:
* Fix typo
* Add proposed user object
* Begin webhooks documentation outlining how time keeping can be automated using github.
* Address #169

### DIFF
--- a/docs/source/design/draft_models.rst
+++ b/docs/source/design/draft_models.rst
@@ -61,5 +61,9 @@ User:
 
 .. code-block:: python
 
-    # This is currently unplanned. We will address this in a future meeting.
-    id # auto-incrementing primary key
+    id, # auto-incrementing primary key
+    first_name, # string
+    last_name, # string
+    github_username, # string
+    password, # hash
+    role, # string (ie. developer, system administrator, writer, etc.)
--- a/docs/source/dev/development.rst
+++ b/docs/source/dev/development.rst
@@ -50,7 +50,7 @@ And run the server::
 
     npm run devel
 
-.. note:: ``npm devel`` uses the ``nodemon`` tool to automatically restart your
+.. note:: ``npm run devel`` uses the ``nodemon`` tool to automatically restart your
     test server when files are changed. To run TimeSync in production, use
     ``npm start``.
 
--- a/docs/source/dev/development.rst
+++ b/docs/source/dev/development.rst
@@ -48,7 +48,7 @@ run the migrations::
 
 And run the server::
 
-    npm devel
+    npm run devel
 
 .. note:: ``npm devel`` uses the ``nodemon`` tool to automatically restart your
     test server when files are changed. To run TimeSync in production, use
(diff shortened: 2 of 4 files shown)


lucywyman:
Potential issues:
* What happens if a project has multiple repos?  This design assumes that a project name can be extracted from the github issue payload as the repository the issue is made on.
* I'm not sure if current_project and current_issues are necessary

lucywyman:
@osuosl/devs  Take a look when you get the chance

docs/source/dev/development.rst
```diff
@@ -48,7 +48,7 @@ run the migrations::
 
 And run the server::
 
-    npm devel
+    npm run devel
 
 .. note:: ``npm devel`` uses the ``nodemon`` tool to automatically restart your
```
iankronquist: You might as well fix it here too.

lucywyman: Derp -- good call.

docs/source/design/draft_models.rst
```diff
@@ -62,4 +62,13 @@ User:
 .. code-block:: python
 
     # This is currently unplanned. We will address this in a future meeting.
-    id # auto-incrementing primary key
+    id, # auto-incrementing primary key
+    first_name, # string
+    last_name, # string
+    username, # string
+    password, # hash
+    role, # string (ie. developer, system administrator, writer, etc.), not sure if nessecary
```
Kennric: In fact we discussed this at the summit, it would be useful for filtering by student vs staff, for instance

docs/source/design/draft_models.rst
```diff
@@ -62,4 +62,13 @@ User:
...
+    current_issues, # Are these the same as activities?
+    # A list of issues the user is currently working on
+    # Alternatively, this could be current projects
+    current_projects,  # List of foreign keys
```
Kennric: We aren't really tracking issues in timesync, we're just storing a record of time spent. A user will be associated with a project (which has a URI, perhaps the main ticket tracker), and each time entry can refer to an external issue URI (an rt ticket, github issue, or something else). The activity is something like "review", "documentation", "coding", etc, it is just a text descriptor. 

So the data these fields seem intended to track are implicit already in the database. Can you provide the user story or use case you have in mind for these fields?

Kennric:
I think what would help me understand the context of these changes are user stories that show how we would use the new fields in the user model.

lucywyman:
@Kennric  I gave user stories a try -- I'm not proud of it, but it is what it is.  I also didn't address auth with github at all -- I'll likely leave that for someone else :)  Hope it helps!

Kennric:
I think the documentation on git webhooks and the user stories are good, but I am not sure I want to commit a change to the draft user model just yet - if you can remove that, I will merge the documentation pieces.

LyonesGamer:
Addressed by https://github.com/osuosl/gitsync.

PR #191, opened 2015-10-06
Transactions for tests!

tschuy:
This makes things fast!

```
  209 passing (6s)
```


tschuy:
Ok, this is ready for review. 

This removes the ability to use sqlite as a database backend, which, given some of the bugs I found, isn't a bad idea.

To run tests, you'll need to make sure your ``PG_CONNECTION_STRING`` connects to an empty database that has had migrations run. This is the same string as the dev uses, so I'd switch between database names before running tests!

Kennric:
Um, that's discomfiting - can the test runner set that variable? That looks like a recipe for losing data otherwise

tschuy:
Ok, I've made a few changes:

1. now it supports both sqlite and postgres backends. to run sqlite tests, run ``npm test``; to run postgres tests, run ``npm run test_pg``.

2. it allows you to specify a separate database connection for tests -- ``TEST_PG_CONNECTION_STRING`` is what is used now.

should be good to go now?

knexfile.js
```diff
@@ -6,13 +6,23 @@ module.exports = {
...
+  development_pg: {
+    client: 'pg',
+    connection: process.env.PG_CONNECTION_STRING
+  },
+
+  mocha_sqlite: {
     client: 'sqlite3',
     connection: {
-      filename: ':memory:'
+      filename: './test.sqlite3'
```
Kennric: Why not in memory?

tschuy: ah, oops. my mistake.

LyonesGamer:
+1

PR #193, opened 2015-10-07
Soft-deleting times

ElijahCaine:
These are the initial tests for soft-deleting times (DELETE ``/times/:uuid``)

### DIFF
--- a/src/times.js
+++ b/src/times.js
@@ -805,7 +805,7 @@ module.exports = function(app) {
     .then(function(time) {
       if (!time) {
         const err = errors.errorObjectNotFound('uuid', req.params.uuid);
-        return res.status(err.status).send(err)
+        return res.status(err.status).send(err);
       }
 
       knex.transaction(function(trx) {
@@ -814,7 +814,7 @@ module.exports = function(app) {
         .then(function(numObj) {
           if (numObj >= 1) {
             trx.commit();
-            return res.send()
+            return res.send();
           }
 
           trx.rollback();
--- a/src/times.js
+++ b/src/times.js
@@ -793,4 +793,11 @@ module.exports = function(app) {
       return res.status(err.status).send(err);
     });
   });
+
+  app.delete(app.get('version') + '/times/:uuid', function(req, res) {
+    const kenx = app.get('knex');
+    if (!helpers.validateUUID(req.params.uuid)) {
+      const err = errors.errorInvalidUUID('uuid', req.params.slug);
+      return res.status(err.status).send(err);
+    }
 };
--- a/src/times.js
+++ b/src/times.js
@@ -809,14 +809,12 @@ module.exports = function(app) {
       }
 
       knex.transaction(function(trx) {
-        // Delete time after running through checks.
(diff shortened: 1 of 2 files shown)


tests/times.js
```diff
@@ -3283,4 +3283,56 @@ module.exports = function(expect, request, baseUrl) {
...
+              created_at: '2015-04-19',
+              updated_at: null,
+              deleted_at: null,
+              uuid: '32764929-1bea-4a17-8c8a-22d7fb144941',
+              revision: 1,
+              id: 1,
+            },
+          ];
+
+          // TODO: GET should only return 200 when ?revisions=true is passed.
```
Kennric: Shouldn't this test return 404, since we deleted the object?

Kennric:
That looks better. +1

LyonesGamer:
I would also add a test for sending an invalid UUID (i.e. a string that can't even be parsed as a UUID), which would return an Invalid Identifier error instead of an Object Not Found.

LyonesGamer:
+1 Looks good to me.

Kennric:
I think that's got it. +1

src/times.js
```diff
@@ -793,4 +793,45 @@ module.exports = function(app) {
...
+    knex('times').select('id').where('uuid', req.params.uuid).first()
+    .then(function(time) {
+      if (!time) {
+        const err = errors.errorObjectNotFound('uuid', req.params.uuid);
+        return res.status(err.status).send(err);
+      }
+
+      knex.transaction(function(trx) {
+        trx('times').where('uuid', req.params.uuid)
+        .update({'deleted_at': Date.now(), 'uuid': null})
```
LyonesGamer: Why are we setting ``uuid`` to null?

src/times.js
```diff
@@ -793,4 +793,45 @@ module.exports = function(app) {
...
+
+    knex('times').select('id').where('uuid', req.params.uuid).first()
+    .then(function(time) {
+      if (!time) {
+        const err = errors.errorObjectNotFound('uuid', req.params.uuid);
+        return res.status(err.status).send(err);
+      }
+
+      knex.transaction(function(trx) {
+        trx('times').where('uuid', req.params.uuid)
```
LyonesGamer: I think we only want to be setting ``deleted_at`` on the "current" revision.

src/times.js
```diff
@@ -793,4 +793,45 @@ module.exports = function(app) {
...
+
+    knex('times').select('id').where('uuid', req.params.uuid).first()
+    .then(function(time) {
+      if (!time) {
+        const err = errors.errorObjectNotFound('uuid', req.params.uuid);
+        return res.status(err.status).send(err);
+      }
+
+      knex.transaction(function(trx) {
+        trx('times').where('uuid', req.params.uuid).first()
```
LyonesGamer: Because order isn't guaranteed on a database call, you should impose order here, ``.orderBy('revision', 'desc')``

LyonesGamer:
+1

Kennric:
Looks good. +1

Issue #201, opened 2015-10-16
Authentication Schemes


LyonesGamer:
Currently, we provide two authentication schemes: password (local password store) and LDAP, which is basically a password auth with a non-local lookup.

We want to move to a slightly more robust scheme, that also allows us to secure GET and DELETE endpoints. We also want to make sure that this is usable by both web and CLI clients.

My current idea is to have some sort of /login endpoint, which you POST to to retrieve a session token. This token is stored by the app locally somehow: for example, in a web app it would store as a secure cookie. This token will also be somehow constructed, reversibly, from a username and creation time (e.g. something like ``creation_time + "$" + username + "$" + hash(salt + username + password)``). It will be valid with the API for some amount of time (likely 30 minutes) and for a single IP. When the client wishes to make a request, it sends the token either as a query parameter (e.g. ``GET /times?token=...``) or in the body (e.g. ``POST /projects {"auth": {"type": "token", "token": "..."}, "object": {...}}``), as well as in the cookie. The server then checks that the token is valid, and has the creation time to check that it hasn't expired, and the username to perform authorization.

I'm still working out the details. Also, we need to look into ideas like OAuth2 (for web) or SSH Keys (for CLI). Thoughts?

matthewrsj:
Your token idea makes sense. From my end, including the token in the body would be easier I think, since it is closer to `password` and `LDAP` auth types. Less code anyway.

LyonesGamer:
Sorry, maybe this wasn't clear. You'd include it as a query parameter on GET and DELETE requests, and in the body for POST requests (in the body is ideal, but GET and DELETE aren't supposed to have bodies).

matthewrsj:
ah, ok

LyonesGamer:
Token authentication has been implemented and merged. Closing this issue.

PR #203, opened 2015-10-30
Implementation of token auth for all endpoints.

LyonesGamer:
Replaces password and ldap authentication on all endpoints with token auth, adds a /login endpoint which uses password or ldap authentication to generate a short-term token for use on other endpoints. Completes #201.


ElijahCaine:
After a pretty thorough review I'd say this all looks good.
mfw the diff is ~7k lines
Once the diff gets resolve and we get a few more eyes on this it should be good to go.

ElijahCaine:
lgtm :+1:

src/login.js
```diff
@@ -0,0 +1,138 @@
+'use strict';
+const errors = require('./errors');
+const passport = require('passport');
+const crypto = require('crypto');
+
+const tokens = {};
+
+const MAX_AGE = 30 * 60 * 1000;
```
Kennric: This should, at some point, be a configurable number

Kennric:
I can't see any issues, please create a new issue to make token age configurable. +1

PR #204, opened 2015-11-02
Thai/include deleted activities

athai:
Add 'include_deleted' parameter to the activities endpoint.

### DIFF
--- a/src/activities.js
+++ b/src/activities.js
@@ -8,7 +8,17 @@ module.exports = function(app) {
 
   app.get(app.get('version') + '/activities', function(req, res) {
     const knex = app.get('knex');
-    knex('activities').whereNull('deleted_at').then(function(activities) {
+
+    // Only query for activities that have not been soft-deleted
+    let activitiesQ = knex('activities').whereNull('deleted_at');
+
+    // If the include_deleted param is true, query for all activities
+    // regardless of 'deleted' status
+    if (req.query.include_deleted) {
+      activitiesQ = knex('activities');
+    }
+
+    activitiesQ.then(function(activities) {
       if (activities.length === 0) {
         return res.send([]);
       }
--- a/src/activities.js
+++ b/src/activities.js
@@ -8,14 +8,15 @@ module.exports = function(app) {
 
   app.get(app.get('version') + '/activities', function(req, res) {
     const knex = app.get('knex');
+    let activitiesQ;
 
-    // Only query for activities that have not been soft-deleted
-    let activitiesQ = knex('activities').whereNull('deleted_at');
-
-    // If the include_deleted param is true, query for all activities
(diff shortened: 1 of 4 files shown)


athai:
Question: If a user tries to specify the deleted activity by 'activities/:slug?include_deleted=true', what error should be returned? I'm going with Object Not Found for now, but is there a better alternative?

ElijahCaine:
@athai Object Not Found is the correct error afaik.

LyonesGamer:
IMO Object Not Found is the best here.

athai:
@ElijahCaine @LyonesGamer Cool, thanks. Also, how do my tests look?

tests/activities.js
```diff
@@ -347,6 +347,158 @@ module.exports = function(expect, request, baseUrl) {
...
+        ];
+
+        expect(getErr).to.equal(null);
+        expect(getRes.statusCode).to.equal(200);
+        expect(jsonBody).to.deep.equal(expectedResults);
+        done();
+      });
+    });
+
+    it('returns an error if user specifies with /activities/:slug endpoint',
```
ElijahCaine: This should describe that the slug does not exist. Sending /activities/ a :slug is valid.

ElijahCaine:
:thumbsup: looks good to me.

LyonesGamer:
+1

src/activities.js
```diff
@@ -8,7 +8,17 @@ module.exports = function(app) {
...
   app.get(app.get('version') + '/activities', function(req, res) {
     const knex = app.get('knex');
-    knex('activities').then(function(activities) {
+
+    // Only query for activities that have not been soft-deleted
+    let activitiesQ = knex('activities').whereNull('deleted_at');
+
+    // If the include_deleted param is true, query for all activities
+    // regardless of 'deleted' status
+    if (req.query.include_deleted) {
```
ElijahCaine: Is there a reason this is two separate statements and not an if/else? Querying the database twice if the parameter is passed seems like a needless performance hit.

athai: How would you encase this in an if/else?

ElijahCaine: I was thinking something like this: 
```
if (req.query.include_deleted) {
  let activitiesQ = knex('activities');
} else {
  let activitiesQ = knex('activities').whereNull('deleted_at');
}
```

LyonesGamer: Or to revise that slightly,

    let activitiesQ = knex('activities');
    if (req.query.include_deleted) {
      activitiesQ = activitiesQ.whereNull('deleted_at');
    }

Kennric:
Tests pass, and the code looks good to me. Get a second opinion. +1

ElijahCaine:
I agree with @Kennric on this. The code looks great. +1

PR #207, opened 2015-11-04
Thai/include deleted projects

athai:


tests/projects.js
```diff
@@ -79,6 +79,176 @@ module.exports = function(expect, request, baseUrl) {
...
+        expect(res.statusCode).to.equal(200);
+
+        expect(jsonBody).to.deep.equal(expectedResults);
+        done();
+      });
+    });
+
+    it('ignores extra param if user specifies query with a projectslug',
+    function(done) {
+      request.get(baseUrl + 'projects?project=chili&include_deleted=true',
```
Kennric: Is this a valid URI? we would normally query a project by slug with projects/<slug> rather than a parameter - do we actually allow ?project=slug on the projects endpoint?

athai: It isn't.  I'm basing the parameter off - https://github.com/osuosl/timesync-node/pull/204. 

According to our API docs (https://github.com/osuosl/timesync/blob/develop/source/draft_errors.rst#bad-query-value),  nonexistent parameters will be ignored - which is what I'm testing for. Since querying for projects is somewhat similar to querying for times, I thought that the mistake would be easy to make. Is it unnecessary to test for this?

Kennric: I think it is a good test, I just wanted to make sure I knew what was going on.

Kennric:
Looks good, +1

tests/projects.js
athai: I'll add comments. :)

ElijahCaine:
@athai These tests are a bit verbose but they get the job done. +1, implementation should be good if it passes these.

LyonesGamer:
Looks good to me. +1

Kennric:
+1

LyonesGamer:
+1

PR #212, opened 2015-11-20
Thai/include deleted times

athai:
Tests for the include_deleted times parameter have been written.

test_data has been modified to include deleted time entries and associated timesactivities


tests/fixtures/test_data.json
```diff
@@ -268,6 +296,16 @@
       "id": 5,
       "time": "times:3",
       "activity": "activities:1"
+    },
+    {
+      "id": 6,
+      "time": "times:4",
```
Kennric: Should this be associated with time 5, and the next one 6, instead of 4 and 5?

tests/times.js
```diff
@@ -1436,6 +1436,242 @@ module.exports = function(expect, request, baseUrl) {
...
+        id: 5,
+        uuid: 'b6ac75fb-7872-403f-ab71-e5542fae4212',
+        revision: 1,
+        deleted_at: '2015-05-02',
+      },
+      {
+        duration: 12,
+        user: 'patcht',
+        project: ['pgd'],
+        activities: ['meeting'],
```
Kennric: Is this activity correctly associated in the test data?

tests/fixtures/test_data.json
athai: I don't think so. Why would it be associated with time 5 instead?

Kennric:
I think these look pretty good, I'd get a second opinion and then start implementing. +1

ElijahCaine:
After a (quick) review these tests look good and pretty consistent with our other tests. +1

tests/fixtures/test_data.json
LyonesGamer: Specifically, the times table is 1-indexed like all SQL databases, but the fixtures syntax used here is 0-indexed and based on the order it's entered into the DB (i.e. the order listed in the file). It's weird.

Kennric:
This looks pretty good. +1

athai:
You can currently query for deleted times that are specific to a project, activity, or user. Should I also add tests that check whether results are narrowed further by combining parameters? That is, tests that check whether requests such as:
* ``GET /times?user=:user&activity=:activityslug&project=:projectslug&include_deleted`` 
* ``GET /times?user=:user&activity=:activityslug&include_deleted``
*  ``GET /times?user=:user&project=:project&include_deleted`` 

returns the correct result. If so, how many combined parameters should I check for?

src/activities.js
```diff
@@ -13,10 +13,11 @@ module.exports = function(app) {
     /* Only query for activities that have not been soft-deleted
      * If the include_deleted param is true, query for all activities
      * regardless of 'deleted' status */
-    if (req.query.include_deleted) {
+    if (req.query.include_deleted === 'true' ||
+        req.query.include_deleted === '') {
       activitiesQ = knex('activities');
     } else {
-      activitiesQ = knex('activities').whereNull('deleted_at');
+      activitiesQ = knex('activities').where({deleted_at: null});
```
ElijahCaine: Is this different or just more consistent with how we've been doing things?

ElijahCaine:
Tests pass, implementation looks straight forward and is easy to understand. +1 from me.

src/activities.js
athai: Just more consistent with how it's done elsewhere.

Kennric:
Further param combo tests are being added.

Kennric:
I think you have everything tested, but there is a lot of code there, please also have @LyonesGamer or @ElijahCaine look it over. +1

ElijahCaine:
Everything looks good to me. Very thorough. :+1:

PR #213, opened 2015-12-02
Add default activities for projects.

### LINKED ISSUE #106: Allow projects to have a default activity
Add a default activity to projects, such that an admin can specify what activity time will be associated with if time entries do not provide an activity.

This requires allowing null activity fields in time entry submissions, and allowing a default activity to be defined on project creation (and edited as needed).

If no default is set for a project, then there are a couple of options: 
1. a default is assigned (requires at least one activity to be created at time of project creation)
2. a default activity is created (project specific?) and assigned on project creation
3. the default is left null, and time entry submissions are then required to provide an activity (more complex logic)

LyonesGamer:
Projects have an optional ``default_activity`` field. When a time is posted with a project with a default activity but no activities, it is given the default activity. If the time is posted with a project with no default activity and no activities, an error is returned. If the time is posted with activities, the project's default activity is ignored. If the project's default activity is changed, it has no impact at all on the time (even if the time was created with the default activity). Fixes #106.

### DIFF
--- a/src/login.js
+++ b/src/login.js
@@ -137,6 +137,8 @@ module.exports = function(app) {
 
             resolve(payload.sub);
           });
+        } else {
+          return reject({message: 'Unsupported algorithm'});
         }
       });
     },
--- a/src/login.js
+++ b/src/login.js
@@ -138,7 +138,8 @@ module.exports = function(app) {
             resolve(payload.sub);
           });
         } else {
-          return reject({message: 'Unsupported algorithm'});
+          return reject({message: 'Unsupported token crypto-algorithm: ' +
+                                                                  header.alg});
         }
       });
     },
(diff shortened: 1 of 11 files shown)


src/login.js
```diff
@@ -131,6 +131,8 @@ module.exports = function(app) {
 
             resolve(payload.sub);
           });
+        } else {
+          return reject({message: 'Unsupported algorithm'});
```
ElijahCaine: This could be a bit more specific.
```
return reject({message: 'Unsupported token crypto-algorithm: ' + header.alg});
```
for example.

ElijahCaine:
Besides the error message this looks pretty good. Tests are good and they pass so that's a plus.

Kennric:
I think that's got it. +1

ElijahCaine:
+1 let's merge

Issue #214, opened 2015-12-02
User and role management


LyonesGamer:
This encompasses a handful of things:

1. Creation of ``admin`` role, to create projects, activities, and users.
2. Ability to set an initial admin user when the instance is first created.
3. Basic CRUD on a ``/users`` endpoint for admin users.
4. ~~Remove deprecated ``owner`` field on projects and set creator as project manager.~~
5. Ability of admins and project managers to designate other users as project members, spectators, and managers.
6. The ``GET /projects`` endpoints should return a list of members, spectators, and managers (or this could be put in a separate endpoint, e.g. ``GET /projects/:slug/members``)

LyonesGamer:
Superseded by #236, #237, #238, #239, #240, #241, #242, https://github.com/osuosl/timesync/issues/58, and https://github.com/osuosl/timesync/issues/59.

PR #215, opened 2015-12-02
Remove project owner, replace with manager in full.

LyonesGamer:
Completes part 4 of #214.


Kennric:
I think that's pretty complete, but another set of eyes would be a good idea. +1

ElijahCaine:
Tests pass, changes make sense. Can't think of anything new that would drastically help or anything in the implementation that will cause problems. LGTM :+1:

PR #217, opened 2015-12-07
Include Revisions on the Projects endpoint

ElijahCaine:
The tests related to /projects/*?include_revisions pass now.
A few other tests (once merged, the include_revisions on the times endpoint for instance) do not pass yet, that's why this merge is going into ``elijah/includ_revisions`` and not ``develop``.

### DIFF
--- a/src/projects.js
+++ b/src/projects.js
@@ -7,89 +7,147 @@ module.exports = function(app) {
   const authPost = require('./authenticatedPost');
   const uuid = require('uuid');
 
+  function constructProject(inProject, slugs) {
+    // manually create our project object from inProject.
+    const outProject = {id: inProject.id, name: inProject.name,
+      owner: inProject.owner, uri: inProject.uri, uuid: inProject.uuid,
+      revision: inProject.revision, created_at: inProject.created_at,
+      updated_at: inProject.updated_at, deleted_at: inProject.deleted_at};
+    if (slugs) { outProject.slugs = slugs.sort(); }
+
+    outProject.created_at = new Date(parseInt(outProject.created_at, 10))
+    .toISOString().substring(0, 10);
+    if (outProject.updated_at) {
+      outProject.updated_at = new Date(parseInt(outProject.updated_at, 10))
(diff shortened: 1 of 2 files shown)


src/projects.js
```diff
@@ -7,53 +7,105 @@ module.exports = function(app) {
...
 
-      // only return the project once both
-      // users and slugs have finished processing
-      let usersDone = false;
-      let slugsDone = false;
+  app.get(app.get('version') + '/projects', function(req, res) {
+    const knex = app.get('knex');
+    const projectsQuery = knex('projects').select(
+      // select the owner to be the users username
+      'users.username as owner',
```
Kennric: Since this issue was created, we've deprecated the owner field - I am not sure where else code may depend on it, but if we can leave it out here, that would be nice. Otherwise merge this and immediately refactor to remove it in a new issue.

src/projects.js
```diff
@@ -99,97 +156,94 @@ module.exports = function(app) {
...
+    .leftOuterJoin('projectslugs', 'projects.id', 'projectslugs.project')
+    // Matching on the uuid subquery (comments above)
+    .where({'uuid': uuidSubquery});
+
+    // The 'include_revisions' query  parameter was included
+    if (req.query.include_revisions === 'true' ||
+        req.query.include_revisions === '') {
+      projectsQuery.then(function(projectsList) {
+        if (projectsList.length !== 0) {
+          // This blog generates the list of slugs
```
Kennric: blog?

Kennric:
Overall, looks pretty good, but we need to make a decision on how to handle the owner field.

src/projects.js
ElijahCaine: Yeah, I think I got distracted while I was typing. Good catch.

ElijahCaine:
I am going to merge this into elijah/include_revisions *with* the owner field and before I merge include_revisions into develop I'll get rid of the owner field. Doing it all at once will help me avoid missing glaring problems.

PR #228, opened 2015-12-22
Remove remaining references to owner.

LyonesGamer:
Fix some code that had been missed previously in #215.

### DIFF
--- a/src/projects.js
+++ b/src/projects.js
@@ -43,34 +43,6 @@ module.exports = function(app) {
         }
       }
 
-      // only return the project once both
-      // users and slugs have finished processing
-      let usersDone = false;
-      let slugsDone = false;
-
-      knex('users').then(function(users) {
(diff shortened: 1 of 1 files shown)


matthewrsj:
You should revise the tests to actually test the right things, since they were passing for #215.

Kennric:
This cleanup looks good, but I agree with @matthewrsj, this should include tests that only pass after this is merged.

LyonesGamer:
The tests passed in #215 because the functionality is identical between the two branches. This one is mostly cleaning up.

The first chunk in ``GET /projects`` is probably adding a lot of extra processing time (especially since it does a knex call for all users), but since there is no owner field, it doesn't actually end up changing the returned data at all.

The rest is mostly comments and a property declaration that does nothing.

But from the perspective of the tests, the code hasn't changed. The data being sent back is the same.

PR #229, opened 2016-01-12
Fix "cannot set headers after they are set" error by cleaning up transactions.

LyonesGamer:
Calling rollback() on a transaction causes the catch function on ``knex.transaction()`` to always be called. By removing the server error blocks from the inner catch blocks, such that these only contain calls to rollback(), we can prevent the "cannot set headers after they are set" error that occurs due to attempting to send an error twice.

### DIFF
--- a/src/activities.js
+++ b/src/activities.js
@@ -119,23 +119,11 @@ module.exports = function(app) {
           // delete activity after running through the checks
           trx('activities').where('slug', req.params.slug)
           .update({'deleted_at': Date.now(), 'slug': null})
-          .then(function(numObj) {
-            /* When deleting something from the table, the number of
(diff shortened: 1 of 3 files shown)


ElijahCaine:
@LyonesGamer What issue / problem does this PR address?

ElijahCaine:
:+1: lgtm, was explained in meatspace. The old code was throwing two errors, which resulted in a ``cannot set headers after response has been sent``. We will implement logging to avoid the inevitable "This is failing but we can't find the errrors... TimeSync... Timesync pls..." in the future.

Kennric:
This looks good - it might be worth thinking of a test that could have uncovered this issue. +1

PR #230, opened 2016-01-19
Add console and file logging.

LyonesGamer:
This will prevent failures to solve problems on the staging (and, in the future, production!) instance by providing full access and error logging on the application, including timestamps, etc.

### DIFF
--- a/src/activities.js
+++ b/src/activities.js
@@ -5,6 +5,7 @@ module.exports = function(app) {
   const helpers = require('./helpers')(app);
   const authRequest = require('./authenticatedRequest');
   const uuid = require('uuid');
+  const log = app.get('log');
 
   function constructActivity(inActivity, res) {
     if (!inActivity) {
@@ -88,6 +89,7 @@ module.exports = function(app) {
           return constructActivity(activity, res);
         }));
       }).catch(function(error) {
+        log.error(req, 'Error requesting activities: ' + error);
         const err = errors.errorServerError(error);
         return res.status(err.status).send(err);
       });
@@ -125,6 +127,7 @@ module.exports = function(app) {
         // Return the frist element since this is a 1 element array
         }).pop());
       }).catch(function(error) {
+        log.error(req, 'Error requesting activity by slug: ' + error);
         const err = errors.errorServerError(error);
         return res.status(err.status).send(err);
       });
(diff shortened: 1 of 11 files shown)


src/log.js
```diff
@@ -0,0 +1,81 @@
...
+
+    if (process.env.DEBUG) {
+      debug = new Log('debug');
+    }
+  }
+
+  setup();
+
+  // Remove tokens from the URL
+  const urlRegex = new RegExp('[?&]token=[a-zA-Z0-9\.=+/]+', 'i');
```
Kennric: Should this be inside the stripUrl method?

src/login.js
```diff
@@ -8,8 +8,14 @@ const tokens = {};
 const MAX_AGE = 30 * 60 * 1000;
 
 module.exports = function(app) {
+  const log = app.get('log');
   let authType;
 
+  if (!process.env.INSTANCE_NAME || !process.env.SECRET_KEY) {
+    log.error('login.js', 'INSTANCE_NAME or SECRET_KEY not set!');
+    process.exit(1);
+  }
```
Kennric: I have questions about this test. 
First, I think any configuration options that are required by the application should be checked when the instance starts, I am not sure login is the place for that test.

Second, if the application is configured by file (i.e. not by setting environment variables), will these variables be in the process.env at all?

src/activities.js
```diff
@@ -228,9 +235,14 @@ module.exports = function(app) {
 
         return res.send(activity);
       }).catch(function(error) {
+        log.error(req, 'Error updating activity: ' + error);
         const err = errors.errorServerError(error);
         return res.status(err.status).send(err);
       });
+    }).catch(function(error) {
```
Kennric: previously there was no catch here at all?

src/log.js
LyonesGamer: I believe not, because if it were, it would recompile the regex each time stripUrl is called, which is a waste. Having it outside means it only compiles once.

src/login.js
LyonesGamer: As it stands, it is checked when the application starts, although I do agree it might make sense to move it to app.js instead of login.js.

No, but we currently don't support configuration via file anyway, so I figured that'd be something to add to this check once we support file configuration.

src/activities.js
LyonesGamer: Correct. It had been missed somehow, previously.

ElijahCaine:
@LyonesGamer Is this supposed to have tests? Why don't we have tests for this?

LyonesGamer:
After some discussion, it would seem that logging does not need to be tested.

ElijahCaine:
Since logging does not need tests, then +1 from me.

Kennric:
+1

Kennric:
+1

PR #206, opened 2015-11-02
?include_revisions=:bool parameter 

ElijahCaine:
resolves #198 when merged

PR #232, opened 2016-01-28
Fix small bugs introduced when merging branches earlier.

LyonesGamer:
As a result of rebasing #206, two small bugs were introduced. 1) A function which used transactions did not call ``commit()`` at the end, which would have prevented the database from being updated. 2) Two functions use the root knex object inside of a transaction for a select call, which one might think appears to be legal, but in fact will cause the application to hang indefinitely. This fixes those bugs.

### DIFF
--- a/docs/source/design/libraries.rst
+++ b/docs/source/design/libraries.rst
@@ -41,27 +41,27 @@ Knex uses Bluebird promises. For information on promises in general, see the
 `MDN documentation`_ on promises. For how to use Bluebird promises specifically,
 see `Bluebird's README`_.
 
-A very important thing to note is knex's `connection pooling`_. Connections to databases
-such as PostgreSQL and MySQL default to a minimum of two pooled connections and a maximum
-of eight. This is certainly sufficient for production at this time, but it may become
-necessary in the future to increase this in order to scale the application.
+.. important::
 
-SQLite, however, is only allowed one connection when using a file, due to
-file access issues; when using an in-memory database, on the other hand, it is allowed
-near-unlimited connections. This means that when running the application locally, which is
-done in a SQLite database, using the root knex object inside of a transaction WILL result
-in deadlock and thus in the application hanging (see `this example of it failing`_;
-`this PR`_ may fix the issue when it is merged, causing overuse of connections to instead
-return an error).
+    Knex maintains `connection pools`_ to its databases; databases such as PostgreSQL and
+    MySQL default to a **minimum of two** pooled connections and a maximum of eight.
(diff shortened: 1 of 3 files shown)


ElijahCaine:
@LyonesGamer Does this actually resolve the issues you were talking about? Do we have tests that check this solution is correct?

Kennric:
+1

docs/source/design/libraries.rst
```diff
@@ -41,9 +41,29 @@ Knex uses Bluebird promises. For information on promises in general, see the
 `MDN documentation`_ on promises. For how to use Bluebird promises specifically,
 see `Bluebird's README`_.
 
+A very important thing to note is knex's `connection pooling`_. Connections to databases
```
ElijahCaine: Just put this in an rst directive:
```
.. important::

    knex's `connection pooling`_ creates connections to databases such as PostgreSQL and MySQL [...]
```

docs/source/design/libraries.rst
```diff
@@ -41,9 +41,29 @@ Knex uses Bluebird promises. For information on promises in general, see the
...
+such as PostgreSQL and MySQL default to a minimum of two pooled connections and a maximum
+of eight. This is certainly sufficient for production at this time, but it may become
+necessary in the future to increase this in order to scale the application.
+
+SQLite, however, is only allowed one connection when using a file, due to
```
ElijahCaine: Tighten up this paragraph. There's only ~2 sentences of important information that the system manager needs to be aware of, try to get those ideas across in as few words as possible. Maybe throw in some **bold important stuff** for good measure.

docs/source/design/libraries.rst
```diff
@@ -41,9 +41,29 @@ Knex uses Bluebird promises. For information on promises in general, see the
...
+file access issues; when using an in-memory database, on the other hand, it is allowed
+near-unlimited connections. This means that when running the application locally, which is
+done in a SQLite database, using the root knex object inside of a transaction WILL result
+in deadlock and thus in the application hanging (see `this example of it failing`_;
+`this PR`_ may fix the issue when it is merged, causing overuse of connections to instead
+return an error).
+
+The fact that testing occurs on an in memory database means that this issue is subverted
```
ElijahCaine: ~~The fact that~~ Because testing occurs on an in-memory database [...]

ElijahCaine:
@LyonesGamer Everything looks mostly good. A few pieces of feedback, after those are corrected +1 merge it.

ElijahCaine:
+1 lgtm

PR #249, opened 2016-03-07
Fix authentication command in README.

LyonesGamer:
Follow-up on #247, fix a minor typo in one line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ If you have a user instance already, then simply first authenticate (italics
 indicate user input):
 
 ```
-$ curl -XPOST -d '{"auth":{"type":"password","_username_":"user","password":"_pass_"}}'
+$ curl -XPOST -d '{"auth":{"type":"password","_username_":"user","password":"_pass_"}}' -H "Content-Type: application/json" http://localhost:8000/v1/login
 ```
 
 This will print out a long JWT token, which you can then place into the data on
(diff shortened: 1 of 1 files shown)


ElijahCaine:
+1

Kennric:
+1

README.md
```diff
@@ -76,7 +76,7 @@ If you have a user instance already, then simply first authenticate (italics
 indicate user input):
 
 ```
-$ curl -XPOST -d '{"auth":{"type":"password","_username_":"user","password":"_pass_"}}'
+$ curl -XPOST -d '{"auth":{"type":"password","_username_":"user","password":"_pass_"}}' -H "Content-Type: application/json" http://localhost:8000/v1/login
 ```
```
iCHAIT: @LyonesGamer 

I feel that this part of code should be like this - 

> If you have a user instance already, then simply first authenticate (angular brackets indicates user input):

```
$ curl -XPOST -d '{"auth":{"type":"password","username":"<user>","password":"<pass>"}}'
```

While setting up,  I was actually having this particular issue, I think this part is a bit confusing as `italics` is not actually getting rendered in the code segment.

Thoughts?

LyonesGamer:
Yeah, I didn't think about the fact that formatting would be escaped inside of a code block. I've changed it to use angle brackets instead.

PR #251, opened 2016-03-14
Added a postgres docker testing container command

ElijahCaine:
So we had a bit of difficulty with the /users endpoint and postgres.

To hopefully sidestep this problem later on I added a command:

  `npm run test_pg_docker`

This will automatically spin up a docker container with postgres
installed. It will then spin up a database and run the postgres-backed
tests suite.

Notes:
  - You should be able to run docker as a non-super-user for this to run
    correctly. (add yourself to the docker group)
  - The docker daemon has to be running.
  - The script has a built in 10 second waiting period between spinning
    up the container and moving on. This is so postgres has ample time
    to create your database. If you get errors related to your table not
    being created you'll probably want to bump that time up a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,18 @@ $ curl -XGET -s localhost:8000/v1/times?token=<token> | python -m json.tool
 
 Your output should look something like the above.
 
+If you would like to test the Postgres database, but don't want to setup a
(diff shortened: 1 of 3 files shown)


LyonesGamer:
Looks good to me. +1

Kennric:
Works well. +1

PR #254, opened 2016-03-18
Update dependencies to newest versions.

LyonesGamer:
Only dependencies left not changed are those related to ESLint, because the new updates introduce a large number of new rules designed for ES6, which we aren't prepared to implement at this time.


Kennric:
Well, tests pass, not much else to say. +1

PR #260, opened 2016-03-31
Add errors.send() helper function to prevent "cannot set headers" error.

### LINKED ISSUE #259: Cannot set headers after they are sent...again.
Promises are eternally problematic, it seems. In two places (times.js:267 and times.js:322), rejecting promises while sending errors causes multiple errors to try to be set. Perhaps better to just reject with the error object.

LyonesGamer:
Fixes #259 by checking whether data has already been sent before sending any errors.

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -206,7 +206,7 @@ module.exports = {
   /*
    * Helper function to ensure HTTP status is sent and prevent repeated send
    */
-  sendError: function(error, res) {
+  send: function(error, res) {
     if (!res.finished) {
       return res.status(error.status).send(error);
     }
--- a/src/errors.js
+++ b/src/errors.js
@@ -202,4 +202,15 @@ module.exports = {
     return createError(400, 'Bad Query Value', 'Parameter ' + key +
     ' contained invalid value ' + value);
   },
+
+  /*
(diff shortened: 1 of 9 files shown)


src/errors.js
```diff
@@ -202,4 +202,15 @@ module.exports = {
     return createError(400, 'Bad Query Value', 'Parameter ' + key +
     ' contained invalid value ' + value);
   },
+
+  /*
+   * Helper function to ensure HTTP status is sent and prevent repeated send
+   */
+  send: function(error, res) {
+    if (!res.finished) {
```
tschuy: Sorry, commenting from a point of ignorance of how errors are currently handled. Are errors that occur but aren't able to be sent still recorded in a log, hopefully with some big warning saying, hey, shit happened?

LyonesGamer: If they're a server error, yes. And almost all the time that we get this error, the second error is pretty irrelevant anyway.

ElijahCaine:
Tests pass, code looks good, great idea I only wish we implemented it soooner. :+1:

Kennric:
I don't see a problem with this, but wish we had a clear test case for the problem. +1

PR #275, opened 2016-05-27
Fix crashes on receiving bad json.

### LINKED ISSUE #271: Invalid JSON in POST body causes crashes
ExpressJS's body-parser middleware, in the case of a bad body (in our case, malformed JSON), returns a SyntaxError. As it stands, we do not handle that, causing it to propagate upward and crash the application. Ideally we should catch the error and return a 400 in that case.

LyonesGamer:
Fixes #271.

Adds a middleware that checks for errors, and if there is an error, it returns a 400. Otherwise, it continues as normal.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -38,9 +38,15 @@ app.use(bodyParser.urlencoded({ extended: true }));
 app.set('knex', knex);
 app.set('log', log);
 
+const errors = require('./errors');
+
+/*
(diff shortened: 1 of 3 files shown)


Kennric:
I think this is fine, but it would be nice to return an informative error message

Kennric:
That's better. +1

ElijahCaine:
LGTM +1

Issue #270, opened 2016-05-20
Unexpected server error when updating time


aaroncohen73:
When attempting to update a time, status code 500 is returned with the text "Unexpected server error." This happens both with Pymesync and sending raw JSON with cURL.

Example:
```
>>> ts.update_time(uuid='71bf593d-ceeb-4000-87b7-754fca76d10a', time={'date_worked': '2016-05-19'})
{u'status': 500, u'text': u'Unexpected server error.', u'error': u'Server error'}

$ curl -H "Content-Type: application/json" -X POST -d '{"auth":{"type":"token", "token":"the token"}, "object": {"duration":1200}}' http://timesync-staging.osuosl.org/v0/times/71bf593d-ceeb-4000-87b7-754fca76d10a
{"status":500,"error":"Server error","text":"Unexpected server error."}
```

LyonesGamer:
Not able to take too much of a look right now, but the error message from the logs is:

``Error retrieving time to update: TypeError: Cannot read property 'username' of undefined``

LyonesGamer:
Looks like the issue is that there are no times on the server, so you can't update that time. A patch will be needed to improve the error message.3

aaroncohen73:
@LyonesGamer That definitely wasn't the case when I was testing it. It looks like all the times were deleted off of the server (which must have happened over the weekend). The issue is still there, as I just submitted a new time to the server and attempted to update it via the above methods (substituting the UUID of the newly created time) and I still got the same error.

LyonesGamer:
I am able to reproduce the error on the staging server, but not locally.

@aaroncohen73 What account were you using to update the time? Yours, or the admin user?

LyonesGamer:
I am uncertain what the problem was, but after updating the staging server to develop, it seems to have gone away.

LyonesGamer:
Issue has been rediscovered, confirmed, and more information has been found.

There is a typo in the SQL query, wherein it does ``INNER JOIN projectslugs ON projectslugs.id = times.project`` instead of ``INNER JOIN projectslugs ON projectslugs.project = times.project``.

This means that if the time has a project with an ID which a projectslugs row also has (even if it doesn't match that project), the wrong slug is returned (but it is simply checked to ensure that it exists, then discarded, preventing any issues, and also thus passing tests).

[...]

PR #276, opened 2016-05-31
Check for nonexistent time and project when updating.

LyonesGamer:
A quick solution to ameliorate #270 (and something we should have been doing this whole time *anyway*), but it won't ultimately be the fix.

### DIFF
--- a/tests/times.js
+++ b/tests/times.js
@@ -4127,19 +4127,23 @@ module.exports = function(expect, request, baseUrl) {
      * Without this function you would see this exact code pretty 26
      * times over.
      */
-    function checkPostToEndpoint(done, postObj, expectedResults, error,
+    function checkPostToEndpoint(done, uri, postObj, expectedResults, error,
     statusCode, postBodies) {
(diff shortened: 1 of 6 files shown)


Kennric:
That looks ok to me, simple short-circuit error +1

ElijahCaine:
@LyonesGamer Is there any way to add a regression test for this?

Kennric:
I think those tests are correct, get a second opinion, though. +1

tests/times.js
```diff
@@ -4127,19 +4127,23 @@ module.exports = function(expect, request, baseUrl) {
      * Without this function you would see this exact code pretty 26
      * times over.
      */
-    function checkPostToEndpoint(done, postObj, expectedResults, error,
+    function checkPostToEndpoint(done, uri, postObj, expectedResults, error,
```
ElijahCaine: Why are we passing uri to this function now? iirc it doesn't get used.

ElijahCaine:
Tests look good. +1 from me. Merge it @LyonesGamer

PR #287, opened 2016-06-13
Make activities and projects check only for active times.

### LINKED ISSUE #286: Projects and activities referenced by old times cannot be deleted
If a project or activity is referenced by a time which has been deleted or updated, it still cannot be deleted. Instead, we should only be checking for active (i.e. neither updated nor deleted) times which reference the activity or project.

LyonesGamer:
Fixes #286.

### DIFF
--- a/src/activities.js
+++ b/src/activities.js
@@ -182,7 +182,9 @@ module.exports = function(app) {
                                                                           res);
       }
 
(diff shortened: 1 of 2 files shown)


MaraJade:
+1

Kennric:
+1

Issue #289, opened 2016-06-13
Allow projects to be filtered by users


LyonesGamer:
A ``?user=`` parameter on the ``GET /projects`` endpoint to filter the results by members. Will require docs changes as well as tests and code.

PR #290, opened 2016-06-13
Add user parameter to projects.

LyonesGamer:
fixes issue #289

## Changes in this PR.

- [X] Test the parameter
- [X] Implement the parameter

## Testing this PR.

1. Run ``npm test``.
2. Create a testing instance of TimeSync, then (after authenticating) make a request to ``/projects?user=tschuy``.

### Expected Output.

1. The two new tests should pass.
2. You should see a list only of the projects to which tschuy belongs.

@osuosl/devs

### DIFF
--- a/tests/projects.js
+++ b/tests/projects.js
@@ -199,7 +199,7 @@ module.exports = function(expect, request, baseUrl) {
   });
 
   describe('GET /projects?user=:username', function() {
-    it('returns all times for a user', function(done) {
+    it('returns all projects for a user', function(done) {
       getAPIToken().then(function(token) {
         const username = 'tschuy';
         request.get(baseUrl + 'projects?user=' + username + '&token=' + token,
--- a/tests/projects.js
+++ b/tests/projects.js
@@ -201,13 +201,14 @@ module.exports = function(expect, request, baseUrl) {
   describe('GET /projects?user=:username', function() {
     it('returns all times for a user', function(done) {
       getAPIToken().then(function(token) {
-        request.get(baseUrl + 'projects?user=tschuy&token=' + token,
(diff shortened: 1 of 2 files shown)


LyonesGamer:
On a side note, I think that we should refactor the times tests to look more like this. One list of all times, then use filter() and map() to get the expected results for a given test, instead of redefining the whole list every time. It'll make it easier to maintain and reduce unnecessary redundancy.

ElijahCaine:
These tests look good. To address the side note I would *not* refactor the time tests unless you find yourself working on the time tests anyway.

LyonesGamer:
Agreed.

LyonesGamer:
@ElijahCaine I could use some help. I've left some console.log statements so you can see what I'm talking about, but basically my code is hitting the resolve() call inside of a promise, but never entering the then() function on that promise (Ctrl+F ``console.log('here')`` and ``console.log('there')``).

tests/projects.js
```diff
@@ -198,6 +198,47 @@ module.exports = function(expect, request, baseUrl) {
     });
   });
 
+  describe('GET /projects?user=:username', function() {
+    it('returns all times for a user', function(done) {
```
Kennric: you mean "returns all projects for a user"?

Kennric:
Another note for a tests refactor - we should really replace user/project/activity names with generic names like user1, user2,  project1, project2, (or valid_user_1, invalid_project, etc) rather than have gwm and tschuy all over our test data. A recursive  /tschuy/user_1/ should take care of that pretty easily.

ElijahCaine:
LGTM +1

Kennric:
+1

PR #295, opened 2016-06-23
Use generic names that describe purpose in tests.

### LINKED ISSUE #294: Replace OSL-specific names with generic names.
As it is, our tests contain very OSL-specific names, with users such as `tschuy` and `patcht` and projects such as `pgd` and `gwm`. Replace these names with more generic ones, both for professionalism and ease of reading by non-OSL users/developers.

LyonesGamer:
Fixes #294.

## Changes in this PR.

- [X] Replace usernames with more generic ones describing their roles
- [X] Replace projects with more generic ones describing their forms

@osuosl/devs

### DIFF
--- a/tests/activities.js
+++ b/tests/activities.js
@@ -6,7 +6,7 @@ function copyJsonObject(obj) {
   return JSON.parse(JSON.stringify(obj));
 }
 
-const defaultUsername = 'tschuy';
+const defaultUsername = 'admin1';
 const defaultPassword = 'password';
 
 module.exports = function(expect, request, baseUrl) {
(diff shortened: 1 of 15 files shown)


ElijahCaine:
This is a very welcomed change. I'd like to see us do this for all future projects.

@LyonesGamer Have you considered changing the example code-blocks in the docs as well? If you don't think that's necessary this gets a :+1: from me.

ElijahCaine:
I think those example I was remembering were in osuosl/timesync repo so there's a lot less to change than I expected.

:+1: from me.

tests/activities.js
```diff
@@ -506,7 +506,7 @@ module.exports = function(expect, request, baseUrl) {
     });
 
     it('successfully creates a new activity by a manager', function(done) {
-      getAPIToken('patcht', 'drowssap').then(function(token) {
+      getAPIToken('sManager', 'drowssap').then(function(token) {
```
Kennric: Site_Manager would read better than sManager, I think

Kennric:
This looks pretty good, though I would prefer more verbose and descriptive names. Not critical, but a find/replace would be nice.

Issue #297, opened 2016-07-13
Docker image


aaroncohen73:
It would be useful to create a Docker image for timesync-node and upload it to Docker Hub so that people working on frontends can easily spin up a local instance of timesync-node and test their code locally (See [here](https://github.com/osuosl/timesync-frontend-flask/issues/96)).

aaroncohen73:
The Dockerfile has been moved to the [osl-dockerfiles](https://github.com/osuosl/osl-dockerfiles) repository

PR #302, opened 2016-07-14
Added Dockerfile

aaroncohen73:
fixes issue #297 

## Changes in this PR.

- [X] Created Dockerfile to build Docker images for timesync-node

## Testing this PR.

1. Run `docker build -t timesync-node .`
2. Run `docker run -p 8000:8000 -d timesync-node`
3. Run `curl -XPOST -d '{"auth":{"type":"password","username":"admin","password":"admin"}}' -H "Content-Type: application/json" http://localhost:8000/v0/login`
4. Run `docker ps` to find the ID of the container then use `docker kill` to kill it.

### Expected Output.

`curl` should return a JSON token.

@osuosl/devs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM node:argon
+
+RUN mkdir -p /opt/code
+WORKDIR /opt/code
+COPY . /opt/code
+
+RUN npm install
+RUN npm run migrations
+RUN npm run create-account -- -u admin -p admin
+
+EXPOSE 8000
+
+CMD [ "npm", "run", "devel" ]


ElijahCaine:
Good: When all of the doker commands run without a hitch. The container seems to run the app just fine.
Bad: That curl command returns the following error:
```
{"status":401,"error":"Authentication failure","text":"Incorrect username or password"}
```
Which *is* a valid TimeSync response, it just isn't the one I wanted.

aaroncohen73:
@ElijahCaine I can't replicate the auth error. After running it with `docker run` try doing `docker exec -it <container ID> bash` and `npm run create-account -- -u admin -p admin`, then `exit` and try running the curl command again

aaroncohen73:
Project Dockerfile is now hosted at https://github.com/osuosl/osl-dockerfiles

Closing

PR #312, opened 2016-10-05
Use UUIDs for references between objects instead of IDs.

### LINKED ISSUE #311: Foreign key references use ID, and aren't updated
Times seem to be being "lost", because they reference projects by ID, and thus when projects are updated, the times no longer point to the valid revision. References need to be updated to use UUIDs instead.

LyonesGamer:
Fixes #311.

## Changes in this PR.

- [X] Times reference projects by UUID
- [x] Projects reference activities by UUID
- [x] Updating projects doesn't break times
- [x] Updating activities doesn't break times
- [x] Updating activities doesn't break projects
- [x] Add and test script to migrate existing data to use UUIDs

## Testing this PR.

1. Run `npm test`

### Expected Output.

All tests pass.

@osuosl/devs

### DIFF
--- a/migrations/20150101000000_00.js
+++ b/migrations/20150101000000_00.js
@@ -29,7 +29,7 @@ exports.up = function(knex) {
     table.increments('id').primary();
     table.string('name').notNullable();
     table.string('uri');
-    table.uuid('default_activity').notNullable();
+    table.uuid('default_activity');
     table.bigInteger('created_at').notNullable();
     table.bigInteger('updated_at').defaultTo(null);
     table.bigInteger('deleted_at').defaultTo(null);
--- a/migrations/20150101000000_00.js
+++ b/migrations/20150101000000_00.js
@@ -29,8 +29,7 @@ exports.up = function(knex) {
     table.increments('id').primary();
     table.string('name').notNullable();
     table.string('uri');
-    table.integer('default_activity').references('id').inTable('activities')
-      .onDelete('set null');
+    table.uuid('default_activity').notNullable();
     table.bigInteger('created_at').notNullable();
     table.bigInteger('updated_at').defaultTo(null);
     table.bigInteger('deleted_at').defaultTo(null);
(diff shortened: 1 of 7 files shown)


scripts/update-database-uuid-links.js
```diff
@@ -0,0 +1,84 @@
...
+                times.renameColumn('project_uuid', 'project');
+              }).then(function() {
+                trx.schema.table('projects', function(projects) {
+                  projects.renameColumn('default_activity_uuid',
+                                        'default_activity');
+                }).then(function() {
+                  console.log('Success! Data has been migrated!');
+                  process.exit(0);
+                }).catch(function(error) {
+                  console.error('Error removing times ID field: ' + error);
```
mathuin: Should this error message say projects or times?  It's on the same nesting level as the projects call on line 43.

LyonesGamer: Fixed.

migrations/20150101000000_00.js
```diff
@@ -29,8 +29,7 @@ exports.up = function(knex) {
     table.increments('id').primary();
     table.string('name').notNullable();
     table.string('uri');
-    table.integer('default_activity').references('id').inTable('activities')
-      .onDelete('set null');
+    table.uuid('default_activity');
```
ElijahCaine: Are we editing a migration or creating a new one?  Will editing the migrations file mess up the production database?  If the update-databse-uuid-links.js script accomplishes the 'not breaking stuff' task that should be made clear somewhere.

Kennric:
Given the apparent difficulty of doing the data migration in JS, maybe we should do that out-of-band. 
Here is what I think we should do:

- write an independent data migration script in python 
- test on a local postgres instance loaded with prod or staging db dump
- merge this PR into develop
- update staging instance
- back up staging db
- run data migration on staging
- test staging with climesync
- fix any bugs
- rinse, repeat
- merge develop into master and apply to production

LyonesGamer:
I migrated the data on timesync-staging, then forced it to use this branch. Looks like all these changes do in fact work! Here's a sample of my testing:

https://pastebin.osuosl.org/43176/

PR #316, opened 2017-01-31
Make create endpoints return *_at field and project update return users

### LINKED ISSUE #315: The `POST /type/` endpoints don't all return the *_at fields
According to the docs, the objects returned from all of the "create object" endpoints should contain `created_at`, `updated_at`, and `deleted_at` fields. However, in practice, while the `users` endpoint returns all three, the `projects` and `activities` endpoints only have the `created_at` field, and the `times` endpoint doesn't provide any of the three fields.

Also, projects do not return the `users` field in the response body when updated.

This inconsistency needs to be sorted out.

LyonesGamer:
Fixes #315.

## Changes in this PR.

- [x] Update tests for returning _at fields
- [X] Update tests for returning users on project update
- [x] Return *_at fields from all create endpoints
- [x] Return `users` field from project update endpoint

## Testing this PR.

1. `npm install`
2. `npm run latte`

### Expected Output.

"382 passing (12s)"

@osuosl/devs

### DIFF
--- a/src/activities.js
+++ b/src/activities.js
@@ -445,6 +445,8 @@ module.exports = function(app) {
           // newly created activity
           obj.created_at = new Date(obj.created_at)
           .toISOString().substring(0, 10);
+          obj.updated_at = null;
+          obj.deleted_at = null;
 
           return res.send(JSON.stringify(obj));
         }).catch(function(error) {
--- a/src/projects.js
+++ b/src/projects.js
@@ -765,6 +765,7 @@ module.exports = function(app) {
               project.revision += 1;
               project.created_at = parseInt(project.created_at, 10);
               project.updated_at = Date.now();
+              project.deleted_at = null;
               project.newest = true;
 
               const oldId = project.id;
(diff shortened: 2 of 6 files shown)


aaroncohen73:
When I try to update a project using Climesync, it freezes after entering all of the fields as if it's waiting for the server to respond back to the request that it sent. The first time it happened, this line appeared in "error.log"

`[Wed Feb 01 2017 11:09:05 GMT-0800 (PST)] ERROR ::ffff:127.0.0.1 "/v0/projects/fdsa": Error retrieving old roles: Error: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?`

but it was only logged once despite replicating the freeze multiple times. To replicate it, start up the devel server, connect to it with Climesync, then create a project and attempt to update it using `up`. I'm pretty sure that this bug was introduced in this branch, since the freezing doesn't happen on the develop branch.

LyonesGamer:
@aaroncohen73 Should be fixed now. I accidentally used the root knex object inside of a transaction, which isn't allowed by knex because they're supposed to be different connections but you should only use one connection at a time.

mathuin:
I am not an approved reviewer but I have run the tests and they all pass.

